PR #2, opened 2017-06-17
Fogmodifier handle + LinkedList funcs

vitalexbel:
getFirst() = get(0)
but getFirst is more understandable.
empty() helps, when no need to destroy the list, but need to remove all items

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ local.properties
 [Dd]ebug/
 [Rr]elease/
 x64/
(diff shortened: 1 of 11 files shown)


Frotty:
Thanks for the PR. Looks good so far, but I think the name of the function "empty" should be "clear" which is a more common term.

Also please add hotdoc to the clear function
👍

Frotty:
Thanks :)

PR #3, opened 2017-06-27
trim() fix

vitalexbel:


Frotty:
what does this fix? spaces at the end didn't work? Nice catch.

vitalexbel:
yeah
that trim() function didnt trim spaces at the end.
because charAt(string.lenght()) returns nothing.
because the last char in every string is charAt(string.lenght - 1)

Frotty:
thx

PR #4, opened 2017-07-24
[UnitIndexer] added handlers for onIndex and onDeindex 

IgorSamurovic:
Basically this offers a proper way to address index and deindex events with a deterministic order - 
everything put into onIndex will happen AFTER an unit is indexed
everything put into onDeindex will happen BEFORE an unit is deindexed

also made it possible to call unit.deindex() which deindexes the unit and returns whether an unit was initially indexed to begin with, this is useful in cases where you want control over indexing (like when you use unit recycling systems)

also added overloads for the print function.

### DIFF
--- a/util/Printing.wurst
+++ b/util/Printing.wurst
@@ -25,9 +25,26 @@ public function Loglevel.getTag() returns string
 			return "|cffFB2700error|r"
 	return "?"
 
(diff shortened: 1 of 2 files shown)


Frotty:
Thanks a lot 🌴

PR #10, opened 2017-11-15
Fix for OnUnitEnterLeave

IgorSamurovic:
As it is right now, OnUnitEnterLeave generates a dummy ability that
1) does not meet techtree requirements
2) is visible
3) doesn't actually work

A simple fix:
`@compiletime function generateAbility()
	new AbilityDefinitionDefend(ABILITY_ID)
	..setName("Leave Detect")
	..setEditorSuffix("(OnUnitEnterLeave)")
	..setArtCaster("")
	..setRequirements("")
	..setRace(Race.Unknown)
	..setButtonPositionNormalX(0)
	..setButtonPositionNormalY(-11)
	..setButtonPositionTurnOffX(0)
	..setButtonPositionTurnOffY(-11)`

### DIFF
--- a/util/UnitIndexer.wurst
+++ b/util/UnitIndexer.wurst
@@ -11,6 +11,8 @@ import OnUnitEnterLeave
 	BREAK THIS SYSTEM since it depends on being the only system to use it.
 	Conversely, it is very likely that this system is going to break them, too.
 	Always make sure you only have one system using SetUnitUserData.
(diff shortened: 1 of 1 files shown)


Frotty:
Tauren can you actually answer irc?
Multiple ppl have used the enter/leave without problems before.
Please actually elaborate on the problem.

Also, why did you close it?

PR #11, opened 2017-11-15
Fixing OnUnitEnterLeave

IgorSamurovic:
As it is right now, OnUnitEnterLeave generates a dummy ability that
1) does not meet techtree requirements
2) is visible
3) doesn't actually work

A simple fix:
`@compiletime function generateAbility()
	new AbilityDefinitionDefend(ABILITY_ID)
	..setName("Leave Detect")
	..setEditorSuffix("(OnUnitEnterLeave)")
	..setArtCaster("")
	..setRequirements("")
	..setRace(Race.Unknown)
	..setButtonPositionNormalX(0)
	..setButtonPositionNormalY(-11)
	..setButtonPositionTurnOffX(0)
	..setButtonPositionTurnOffY(-11)`

The 0, -11 value pair for X, Y UI coordinates means the icon is invisible


Frotty:
Define "doesn't work"? I Have used the package successfully several times.
Changes look good and can't hurt - still I'm curious.

IgorSamurovic:
It didn't work as in "the icon was green because wc3 couldn't find the icon file (it being blank), it was visible, and the unit couldn't use it because it required "defend" to be researched"..

I don't know what was up with that, I remember it working previously too, but as it is right now there's no way it works.

Frotty:
As we discovered, it was a problem with your project and not the stdlib.
Please add separate PRs for other stdlib updates.

PR #15, opened 2017-11-21
add public visibility to string methods

IgorSamurovic:
For some reason strings.isUpper and strings.isChar didn't have public acess, so I just added them.

### DIFF
--- a/wurst/_handles/primitives/String.wurst
+++ b/wurst/_handles/primitives/String.wurst
@@ -59,6 +59,8 @@ public function string.toAbilityId() returns int
 	
 /**	Returns if the given string ends with the specified string.	*/
 public function string.endsWith(string s) returns boolean
+	if s == ""
+		return true
 	let len = this.length()
 	if s.length() > len
 		return false
@@ -82,9 +84,6 @@ public function string.toUpperCase() returns string
 /** True when string contains only uppercase letters */
 public function string.isUpper() returns boolean
 	return this == StringCase(this,true)
-
-public function string.ischar() returns boolean
-	return S2I(this) == 0 and this != "0"
 
 /** Returns the given string with whitespaces removed from both ends*/
 public function string.trim() returns string
--- a/wurst/_handles/primitives/String.wurst
+++ b/wurst/_handles/primitives/String.wurst
@@ -2,8 +2,6 @@ package String
 import NoWurst
 import Real
 
-import HashMap
-
 constant charset = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
 constant numberset = "0123456789"
 constant numbersetlength = numberset.length()
--- a/wurst/_handles/primitives/String.wurst
+++ b/wurst/_handles/primitives/String.wurst
@@ -2,6 +2,8 @@ package String
 import NoWurst
 import Real
 
+import HashMap
+
 constant charset = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
 constant numberset = "0123456789"
 constant numbersetlength = numberset.length()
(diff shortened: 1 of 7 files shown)


wurst/_handles/primitives/String.wurst
```diff
@@ -251,6 +253,58 @@ public function string.format(string par1, string par2, string par3, string par4
 				.replace("{8}", par9)
 				.replace("{9}", par10)
 				
+/** Replaces consecutive repeats of the same string with a single instance 
+	Turns "strstr" into "str" */
+public function string.reduce(string reduced) returns string
```
Cokemonkey11: What's the primary use case for this method? Feels like it belongs in a `StringExtensions` or `StringUtils` library, rather than the standard library.

wurst/_handles/primitives/String.wurst
```diff
@@ -251,6 +253,58 @@ public function string.format(string par1, string par2, string par3, string par4
...
+	string toReduce = reduced + reduced
```
Cokemonkey11: use `var`

wurst/_handles/primitives/String.wurst
```diff
@@ -251,6 +253,58 @@ public function string.format(string par1, string par2, string par3, string par4
...
 				.replace("{9}", par10)
 				
+/** Replaces consecutive repeats of the same string with a single instance 
+	Turns "strstr" into "str" */
+public function string.reduce(string reduced) returns string
+	string toReduce = reduced + reduced 
+	string oldString = this
+	string newString
+
+	while true
```
Cokemonkey11: There's an `O(n)` solution for this but not sure how much faster in practice. I think this is fine.

wurst/_handles/primitives/String.wurst
```diff
@@ -251,6 +253,58 @@ public function string.format(string par1, string par2, string par3, string par4
...
+		if newString.length() == oldString.length()
+			break
+		else
+			oldString = newString
+
+	return newString
+
+/** Replaces consecutive repeats of the whitespace character with a single instance 
+	Turns "  " into " " */
+public function string.reduce() returns string
```
Cokemonkey11: As above, what's the primary value of this fn?

wurst/data/LinkedList.wurst
```diff
@@ -223,8 +225,8 @@ class LLEntry<T>
 		this.next = next
 
 public class LLStaticIterator<T>
-	LLEntry<T>    dummy
-	LLEntry<T>    current
+	LLEntry<T>	dummy
```
Cokemonkey11: ??? Never replace spaces with tabs

wurst/data/LinkedList.wurst
```diff
@@ -244,7 +246,7 @@ public class LLStaticIterator<T>
 			current.prev.next = current.next
 			current.next.prev = current.prev
 			let removedNode   = current
-			current           = current.prev
+			current		   = current.prev
```
Cokemonkey11: I see that you fixed mixed whitespace -> tabs only. Much prefer spaces only for visual alignment.

wurst/data/LinkedList.wurst
```diff
@@ -327,6 +329,43 @@ public interface LLItrClosure<T>
 public interface LinkedListUpdater<T>
 	function update(T t) returns T
 
+/*  String processing extra functions 
+	The only reason this is here is that it can't be in String.wurst due to dependency problems */
```
Cokemonkey11: Not a good enough reason. Is there a circular dependency?

wurst/data/LinkedList.wurst
```diff
@@ -327,6 +329,43 @@ public interface LLItrClosure<T>
...
+
+/** Splits the string by another string 
+	"this..is..an..example" split by the string ".." splits into a list containing four strings ["this", "is", "an", "example"] */
+public function string.split(string splitBy) returns LinkedList<string>
+	let list = new LinkedList<string>
+	let offset = splitBy.length()
+	var str = this
+	while true
```
Cokemonkey11: `while str.indexOf(splitBy) != -1`

wurst/data/LinkedList.wurst
```diff
@@ -327,6 +329,43 @@ public interface LLItrClosure<T>
...
+		let first  = str.substring(0, splitIndex)
+		let second = str.substring(splitIndex + offset, str.length())
+		list.add(first)
+		str = second
+	return list
+
+/** Joins elements from a string list into one string using a separator */
+public function LinkedList<string>.joinBy(string separator) returns string
+	var joined = ""
+	let iter = this.iterator()
```
Cokemonkey11: Not sure, but can't you do `for str from this`, which close the iterator automatically? (You're leaking an iterator now, I think. Again, not sure)

wurst/data/LinkedList.wurst
```diff
@@ -327,6 +329,43 @@ public interface LLItrClosure<T>
...
+	for str from iter
+		if iter.hasNext()
+			joined += str + separator
```
Cokemonkey11: I see why you want the iterator now.

May I suggest something more like

```
def list join iter:
    case list.len
        0 => ""
        1 => list.head
        otherwise => {
            var ret = list.head
            let iterator = list.iter
            while iterator.hasNext()
                ret += (iter + iterator.next())
            ret
        }
```

wurst/data/LinkedList.wurst
IgorSamurovic: yes but I figured I will just put it in StringUtils and there's no problem

wurst/_handles/primitives/String.wurst
IgorSamurovic: The primary use for it is when you're templating strings, and have missing values (intentionally), so you would turn "something  something    something" to "something something something", it's a niche use but it's pretty nice to have. I moved it to StringUtils anyways, you're right.

Frotty:
add unit tests for the new functions please

wurst/_handles/primitives/String.wurst
```diff
@@ -251,6 +251,37 @@ public function string.format(string par1, string par2, string par3, string par4
...
+/** Removes all whitespace from the end of a string */
+public function string.rstrip() returns string
+	return this.rstrip(" ")
+
+/** Removes a string from both ends of another string */
+public function string.strip(string val) returns string
+	return this.lstrip(val).rstrip(val)
+
+/** Removes all whitespace from both ends of a string */
+public function string.strip() returns string
```
Frotty: "strip" already exists as "trim".

IgorSamurovic: yes but it only removes whitespaces

wurst/data/LinkedList.wurst
IgorSamurovic: I don't want to call indexOf twice, which is why I did this

Frotty:
Thank you :bowtie:

IgorSamurovic:
```(list.getSize() == 0).assertTrue()```
should be
```list.getSize().assertEquals(1)```

PR #16, opened 2017-11-21
fix broken wording and add .isEmpty and .contains methods

IgorSamurovic:
I also think we should add .isEmpty() to all data structures, what do you think?

### DIFF
--- a/wurst/data/HashList.wurst
+++ b/wurst/data/HashList.wurst
@@ -13,7 +13,6 @@ public class HashList<T>
 	function set(int index, T elem)
 		ht.saveInt(this castTo int, index, elem castTo int)
 
-
 	/** Add all elements from elems to this list */
 	function addAll(HashList<T> elems)
 		for T elem in elems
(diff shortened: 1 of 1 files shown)


wurst/data/HashList.wurst
```diff
@@ -24,27 +23,37 @@ public class HashList<T>
...
 			if t castTo int == ht.loadInt(this castTo int, i)
 				removeAt(i)
 				break
-	
+
 	/** Get the size of the list */
 	function getSize() returns int
 		return size
 
+	/** Return whether the list is empty */
```
Cokemonkey11: What's the value of this comment?

wurst/data/HashList.wurst
```diff
@@ -24,27 +23,37 @@ public class HashList<T>
...
+	function isEmpty() returns bool
+		return size == 0
+
 	/** Get the element at the given index from this list */
 	function get(int index) returns T
 		return ht.loadInt(this castTo int, index) castTo T
 
+	/** Return whether the element exists in the list */
```
Cokemonkey11: How are elements compared? (by value)

References are compared for equality.

Cokemonkey11:
Looks good overall.

Frotty:
good, thanks! 🌴

PR #17, opened 2017-11-22
fix typecasting table not being initialized in compiletime

IgorSamurovic:
I made it a constant and made sure it is initialized right away in compiletime.


IgorSamurovic:
Actually, I just realized that, while this doesn't give errors, it also doesn't work at all. It gives an empty string in compiletime every time!

Frotty:
Your fix wasn't correct. It has been addressed via 238cded

PR #19, opened 2017-12-05
Introduce new Syncing libraries

SamuelMoriarty:
Each library comes with an extensive doc detailing the principles behind the library, the standard usage scenario as seen by the end user, and in the case of Network, an implementation overview to help you get oriented.

There are also a few minor changes in the other packages that I added during the development of this, that I thought deserved to be in the stdlib.


wurst/_handles/Trigger.wurst
```diff
@@ -1,5 +1,13 @@
 package Trigger
 import NoWurst
+import Player
```
Frotty: The changes you made here overlap with stuff from RegisterEvents

wurst/closures/Execute.wurst
```diff
@@ -0,0 +1,32 @@
+package Execute
+
+public interface ForForceCallback
+	function callback()
+
+force dummy = CreateForce()
```
Frotty: You can use `bj_FORCE_ALL_PLAYERS` instead of a dummy

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+			end)
+
+		5. A SimpleSynchronizer can only be used once, and gets destroyed automatically after all players have synced.
+		To synchronize again, create a new SimpleSynchronizer
+			
+********************************************************************************/
+
+// briefly selects the specified unit by the player
+function player.onceSelect(unit what)
+	let temp = CreateGroup()
```
Frotty: use ENUM_GROUP and for from to not have group leaks

wurst/_handles/Player.wurst
```diff
@@ -5,9 +5,11 @@ import Unit
 
 /** Use this array instead of Player() to avoid leaks */
 public player array players
+public player localPlayer
```
Frotty: Can this be a constant?

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+	private static constant integer DUMMY_GHOST  = 'Aeth'
+	private static thistype array reverseLookup
+
+	private unit dummy
+	private bitset syncedPlayers = emptyBitset()
+	private static bitset allPlayers = bitset(4095)
+	private SynchronizationCallback callback = null
+
+	construct()
+		dummy = CreateUnit(DUMMY_PLAYER, DUMMY_ID, 0, 0, 0)
```
Frotty: use cascade ..

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+		dummy.addAbility(DUMMY_GHOST)
+		dummy.pause()
+		dummy.setScale(0)
+		reverseLookup[dummy.getIndex()] = this
+
+	ondestroy
+		reverseLookup[dummy.getIndex()] = null
+		dummy.remove()
+
+	// signals that the local player is ready to continue
```
Frotty: make these hotdoc

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
+package SyncSimple
+
+import MapBounds
+import UnitIndexer
+import ErrorHandling
+import BitSet
+import LinkedListModule
+
+/********************************************************************************
+	SyncSimple library v1.0 by MoriMori.
```
Frotty: probably omit version inside stdlib like other packages

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+		There may be other usages related to async network events as well.
+
+	Usage:
+		1. Create a new SimpleSynchronizer using:
+			let synchronizer = new SimpleSynchronizer()
+
+		2. Send some network events on one (or more) player using the Sync natives (or something else)
+			if sender == localPlayer
+				// value1 and value2 are some local values only known to localPlayer
+				SyncInteger(gc, mkey, vkey1, value1)
```
Frotty: I believe the native is SyncStoredInteger, please clarify parameters

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+				SyncInteger(gc, mkey, vkey1, value1)
+		
+		3. Call .sync() on all players after calling the Sync natives
```
Frotty: "on all players?" is this the line I need, or do I need more

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+			synchronizer.sync()
+
+		4. Set a callback to be called after all players have synchronized and read the data (or do something else)
```
Frotty: make the explanation more general and avoid "(something else)" brackets :)
e.g. Set a callback to be executed when the data has been successfully synchronized between all players

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
+package Network
+
+import SyncSimple
+import Table
+import ErrorHandling
+import MoriUtil
+
+// max amount to be written/read/synced per an execution to avoid hitting the OP limit
```
Frotty: make these comments hotdoc

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
+package Network
+
+import SyncSimple
+import Table
+import ErrorHandling
+import MoriUtil
```
Frotty: MoriUtil?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+	private integer booleanReadIndex = -1
+	private integer stringReadIndex  = -1
+	private integer stringIntReadIndex = -1
+
+	private integer compactIndex = -1
+	private integer explodeIndex = -1
+
+
+	construct()
+		currentMode = WRITING
```
Frotty: can be inlined

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+
+	ondestroy
+		destroy buffer
+
+	private function checkMode(StreamMode toCheck)
+		if currentMode != toCheck
```
Frotty: put `if SAFETY_CHECKS_ENABLED` in here instead of wrapping each call

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+
+import SyncSimple
+import Table
+import ErrorHandling
+import MoriUtil
+
+// max amount to be written/read/synced per an execution to avoid hitting the OP limit
+// this value can be higher if Wurst is compiled without stack traces and with all optimizations turned on
+// and even higher with safety checks off
+@configurable constant integer DATA_PER_EXECUTE = 256
```
Frotty: you should omit the type of constants. perhaps also consider replacing all "integer" with "int"

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+		booleans, strings and reals.
+		If you need data of non-fixed length, use a static packet structure
+		with length fields, like so:
+			var field1 = stream.readInteger()
+			var field2 = stream.readInteger()
+			var length = stream.readInteger()
+			for i = 0 to length - 1
+				someArray[i] = stream.readInteger()
+
+		6. Alternatively, your classes can implement Serializable, like so:
```
Frotty: an alternative shouldn't be "the 6th step". Please provide small, concise, correct examples for certain scenarios and separate them accordingly

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+// characters that can be used as GC keys
+// their amount also determines the size of a single 'round'
+constant string GC_KEYS  = " !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`{|}~"
+
+// precomputed set of GC keys
+string array gcKeys
+// count of GC keys
+integer gcKeysCount = 0
+// lookup table for alphabet
+Table alphabetLookup
```
Frotty: Should be a constant, and can be a HashMap ?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+/**
+	This class is used as an intermediate buffer to store data before it is synchronized.
+	That way, we can send data through gamecache using multiple bursts, while keeping the
+	gamecache key lengths short.
+**/
+public class DataStream
+	// max characters to store per compacting round
+	private static constant integer CHARS_PER_COMPACT = 512
+
+	private Table buffer = new Table()
```
Frotty: can these Tables be replaced by HashMaps?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+	READING
+	LOCKED
+
+/**
+	This class is used as an intermediate buffer to store data before it is synchronized.
+	That way, we can send data through gamecache using multiple bursts, while keeping the
+	gamecache key lengths short.
+**/
+public class DataStream
+	// max characters to store per compacting round
```
Frotty: omit type

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+	private integer currentStringInt  = 0
+	private integer currentRound   = 0
+
+	// finish callback
+	private NetworkFinishedCallback finishCallback
+
+	construct(player sender)
+		this.currentState = NetworkState.PREPARING
+		this.sender = sender
+		this.gcIntegerStream = new GamecacheStream(dataCache, mkey, StreamType.INTEGER)
```
Frotty: can be inlined?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+		// calculate max length in each entity stream
+		let maxLength = max(max(integerCount, realCount), max(booleanCount, stringIntCount))
+
+		// calculate the amount of required sync rounds
+		// we can only send up to gcKeysCount in each sync round, 
+		// so we need to do it multiple times if there's more data than that
+		let syncRounds = maxLength div gcKeysCount + 1
+
+		// save the amount of sync rounds required so that other players know how many times they need to read
+		if localPlayer == sender
```
Frotty: This if and the next one can be merged

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+				metaCache.syncInteger(mkey, getKey(i))
+
+		// send sync event
+		synchronizer.sync()
+		synchronizer.onSynced(() -> receiveMetadata())
+
+	// receives info about the transmission after it has been sent by the originating player
+	private function receiveMetadata()
+		// check that we are transitioning from the correct state
+		if SAFETY_CHECKS_ENABLED and  currentState != NetworkState.SENDING_META
```
Frotty: 2 whitespace after `and`

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+			if localPlayer != sender
+				dataStream.compactStrings()
+			dataStream.setMode(StreamMode.READING)
+			dataStream.resetRead()
+			currentState = NetworkState.FINISHED
+			finishCallback.onFinish(dataStream)
+			// destroy this instance
+			destroy this
+
+	// this is the function to start sending all data in the intermediate dataStream buffer
```
Frotty: make these comments hotdoc

Frotty:
consider renaming **Stream classes as the name implies different functionality/implementation.

wurst/closures/Execute.wurst
SamuelMoriarty: Unfortunately, to execute the callback exactly once, there needs to be only one player inside the Force. I think using bj_FORCE_ALL_PLAYERS would cause it to execute multiple times instead.

wurst/_handles/GameCache.wurst
```diff
@@ -0,0 +1,87 @@
...
+	SyncStoredReal(this, missionKey, key)
+
+public function gamecache.syncBoolean(string missionKey, string key)
+	SyncStoredBoolean(this, missionKey, key)
+
+public function gamecache.syncUnit(string missionKey, string key)
+	SyncStoredUnit(this, missionKey, key)
+
+public function gamecache.syncString(string missionKey, string key)
+	SyncStoredString(this, missionKey, key)
```
Cokemonkey11: Missing EoF LF

wurst/_handles/Player.wurst
```diff
@@ -5,9 +5,11 @@ import Unit
...
 init
 	for i = 0 to bj_MAX_PLAYER_SLOTS-1
 		players[i] = Player(i)
+	localPlayer = GetLocalPlayer()
```
Cokemonkey11: No need to have this in `init` block, I don't think?

If it is needed: needs a comment

wurst/closures/Execute.wurst
Cokemonkey11: use `let` or `constant`, and `dumy` isn't a suitable name

wurst/closures/Execute.wurst
```diff
@@ -0,0 +1,32 @@
...
+
+init
+	ForceAddPlayer(dummy, GetLocalPlayer())
```
Cokemonkey11: Can this just be `..addPlayer(localPlayer)`? If not, can you add _force API?

wurst/closures/Execute.wurst
```diff
@@ -0,0 +1,32 @@
...
+
+function callback() returns boolean
```
Cokemonkey11: not a suitable function name for this internal fn

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
+package Network
+
+import SyncSimple
+import Table
+import ErrorHandling
+import MoriUtil
+
+// max amount to be written/read/synced per an execution to avoid hitting the OP limit
+// this value can be higher if Wurst is compiled without stack traces and with all optimizations turned on
+// and even higher with safety checks off
```
Cokemonkey11: The above two comment are of the form "optimiations, safety checks, and stack traces affect suitable op-limit values". I don't think they add much value here.

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+@configurable constant integer DATA_PER_EXECUTE = 256
+// max length of a GC key
+// you may fine-tune it as you see fit to see what works faster
+@configurable constant integer MAX_CHARACTERS_PER_KEY = 3
+// setting this to false will optimize out all safety checks...
+// only disable this if you are 100% sure your code works correctly
+@configurable constant boolean SAFETY_CHECKS_ENABLED = true
+
+//TODO: I just realized that we can eliminate some delay by sending metadata and the first round together
```
Cokemonkey11: Don't commit TODOs

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+// and avoid one round-trip, essentially reducing transmission time for small payloads up to two times
+
+/********************************************************************************
+	Network library v1.0 by MoriMori.
+
+	Based on TriggerHappy's Sync library.
+
+	Rationale:
+		This library can be used to send arbitrary amounts of integer, real, boolean
+		and string data from one player to the rest.
```
Cokemonkey11: Why would I want to do that?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+
+		each of which can be written/read to using the relevant 
+		methods of the DataStream class.
+
+		Before sending the data, the sender populates the DataStream with the data that
+		they want to send, then Network.start() is called, as well as Network.onFinish(callback)
+		and data is received from the same buffer inside the callback when it has all been received.
+
+		Read SyncSimple docs for a slightly more in-depth overview of this particular
+		peculiarity of WC3.
```
Cokemonkey11: Link?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+
+	Usage:
+		1. Create a new Network object using:
+			// sender is the player to receive data from
+			let network = new Network(sender)
+
+		2. Queue some data for sending from the sender. You can write as much as you need, there is no limit:
+			if localPlayer == sender
+				// keep in mind that before you start sending data,
+				// network.getData() is locked in a WRITE-ONLY MODE
```
Cokemonkey11: Why? What does that mean?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+				val stream = network.getData()
```
Cokemonkey11: s/val/let/

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+				stream.writeInteger(...)
+				stream.writeReal(...)
+				stream.writeString(...)
+				...
+
+		3. To start sending the data, simply call:
+			network.start()			
+			// this will start the transmission
+			// keep in mind, that after you call this, you MUST NOT write/read to network.getData()
+			// as it is locked while the data is syncing
```
Cokemonkey11: How is the stream processed at the receiver? Is there an Rx Queue?

"MUST NOT" - what happens if I try?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+		
+		4. To receive the data, specify the callback and read the data inside it:
+			network.onFinish((stream) -> begin
```
Cokemonkey11: Is this documentation for accepting Tx stream from another host, or a callback function for Tx finished its async operation?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+				// here, network.getData() is locked in a READ-ONLY MODE and
+				// you can only READ from it
+				var int = stream.readInteger()
+				var real = stream.readReal()
+				var string = stream.readString()
+				...
+			end)
+
+			WARNING: Network is a one-time use class, and is automatically
```
Cokemonkey11: You mean: `start()` performs a `destroy this`?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+
+		6. Alternatively, your classes can implement Serializable, like so:
+			class MyClass implements Serializable
+				integer field1 = ...
+				integer field2 = ...
+				integer field3 = ...
+				string field4 = ...
+
+				// optional stream-constructor
+				construct (DataStream stream)
```
Cokemonkey11: is the space between `construct` and `(` standard practice?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+				
+				function deserialize(DataStream stream)
+					field1 = stream.readInteger()
+					field2 = stream.readInteger()
+					field3 = stream.readInteger()
+					field4 = stream.readString()
+
+		And then use it like so:
+			// writing
+			val myClass = ...
```
Cokemonkey11: Not clear. I guess you mean `let myClass = new MyClass(...)` ?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+	Implementation overview:
+		There are 3 core classes here:
+			DataStream
+			GamecacheStream
+			Network
+
+		DataStream is a hashtable-based container with stream semantics for writing
+		integers, reals, booleans, strings and serializables.
+		Because we want to keep gamecache keys below a certain size, we can't store
+		everything immediately in the gamecache without risking exhausting available keys.
```
Cokemonkey11: You mean that `gamecache` might have a collision? Source?

wurst/network/SyncSimple.wurst
Cokemonkey11: https://wurstscript.github.io/manual.html#loops

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+	use LinkedListModule
+
+	private static constant player  DUMMY_PLAYER = players[PLAYER_NEUTRAL_PASSIVE]
+	private static constant integer DUMMY_ID     = 'hfoo'
+	private static constant integer DUMMY_GHOST  = 'Aeth'
+	private static thistype array reverseLookup
+
+	private unit dummy
+	private bitset syncedPlayers = emptyBitset()
+	private static bitset allPlayers = bitset(4095)
```
Cokemonkey11: why `2^12 - 1` ?

wurst/network/SyncSimple.wurst
Cokemonkey11: https://wurstscript.github.io/manual.html#cascade-operator-dot-dot-operator

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+import LinkedListModule
+
+/********************************************************************************
+	SyncSimple library v1.0 by MoriMori.
+
+	Based on TriggerHappy's SyncInteger library.
+
+	Rationale:
+		This library can be used to 'synchronize' players after some async network
+		action has been performed by one or more players.
```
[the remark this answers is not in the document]
SamuelMoriarty: This is perhaps a misnomer, but 'syncing' in WC3 is commonly referring to the act not of recovering from desyncs, but rather sending local data from one player to the rest, so that it can be safely used in synchronous code.

One such example is the various natives that do not take a Player parameter, like GetCameraPos. It is asynchronous and would desync if you used it in synchronous code to get the position of some player's camera. However, with this, you can send one player's camera position to another safely without risking a desync.

Another much more useful example is in save/load systems that work with files. Since file data is only available to one player, we need to sync it to others as well.

When I finish my MultifileIO library, it will be much more obvious! I will also provide various test cases soon for Network library soon when I have the time, that can serve as usage examples.

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+
+		It abuses the fact that all network events in WC3 are fired sequentially,
```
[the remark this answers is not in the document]
SamuelMoriarty: Thanks

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+		meaning that they arrrive for other players in the order that they were
+		sent in.
+		It also abuses the fact that the EVENT_PLAYER_UNIT_SELECTED fires
+		synchronously for all players at the same time, allowing us to know for
```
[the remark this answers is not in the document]
SamuelMoriarty: Deduction based on knowledge about the WC3 game protocol and TCP. So far it holds true. If it doesn't, then I will revert to TH's method of running a periodic timer.

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+		certain when other players have acknowledged our unit selection, as well
+		as all network events that have been fired before it.
+		This is due to the fact that the WC3 game protocol is built on top of TCP,
```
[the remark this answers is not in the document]
SamuelMoriarty: What do you mean?

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,170 @@
...
+
+		By calling the .sync() method, we queue a network action (specifically,
+		a unit selection event) that will only be delivered after all previously 
+		queued network actions have also been delivered.
+		The primary use of this library is in conjunction with gamecache's Sync
+		natives, because they also fire sequential network events. When we call
+		.sync() after a series of Sync natives, we ensure that the .onSynced()
+		callback will only be called after all players have received the data.
+
+		This way, we can send local data from one player to the rest, such as
```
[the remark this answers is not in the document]
SamuelMoriarty: What about the usage section? It provides a valid usage case, which is also the operating principle behind Network.

wurst/network/SyncSimple.wurst
SamuelMoriarty: Thanks

wurst/network/SyncSimple.wurst
SamuelMoriarty: "on all players" in this case means that each player has to call it in order to proceed with the sync. If one or more player doesn't call it, e.g. in a localPlayer clause, then it will not sync.

wurst/network/SyncSimple.wurst
SamuelMoriarty: Thanks

wurst/_handles/GameCache.wurst
SamuelMoriarty: Thanks

wurst/_handles/Player.wurst
SamuelMoriarty: Probably, thanks

wurst/_handles/Player.wurst
SamuelMoriarty: Will inline into a constant

wurst/_handles/Trigger.wurst
SamuelMoriarty: Oh, thanks, did not realize it was there. Thought it'd be intuitive to have those extension functions for Trigger.

wurst/closures/Execute.wurst
SamuelMoriarty: Thanks

wurst/closures/Execute.wurst
SamuelMoriarty: Thanks. Will add _force API as well.

wurst/closures/Execute.wurst
SamuelMoriarty: What name would be more suitable?

wurst/network/Network.wurst
SamuelMoriarty: Old leftover package from initial draft. Will remove.

wurst/network/Network.wurst
SamuelMoriarty: Ok

wurst/network/Network.wurst
SamuelMoriarty: I think these are important considerations for those who may feel like they are experiencing performance issues, and are not necessarily obvious to everyone. Especially not obvious what this constant is for, and why you may want to adjust it, hence, the comment.

wurst/network/Network.wurst
SamuelMoriarty: Haha, yeah, thanks!

wurst/network/Network.wurst
SamuelMoriarty: Read my other long comment ;)

wurst/network/Network.wurst
SamuelMoriarty: Hopefully makes much more sense with file example explanation.

wurst/network/Network.wurst
SamuelMoriarty: Thanks! Habit from writing Kotlin code ;)

wurst/network/Network.wurst
SamuelMoriarty: Frotty and I agreed that 'Stream' is a misnomer. Network is ultimately datagram based and sends the payload in smaller datagrams of `gcKeysCount` size.

If you try with `SAFETY_CHECKS_ENABLED == true`, you will simply receive an error. If you try without them, then it is undefined behaviour. You will probably just screw up the transmission.

wurst/network/Network.wurst
SamuelMoriarty: 'Stream' is a misnomore. This is a callback function for when the datagram has been fully received.

wurst/network/Network.wurst
SamuelMoriarty: `destroy this` is performed after the callback specified in `.onFinish(cb)` is called.

wurst/network/Network.wurst
SamuelMoriarty: Thanks. Will split into different examples.

wurst/network/Network.wurst
SamuelMoriarty: Whoops, my bad

wurst/network/Network.wurst
SamuelMoriarty: Yes. Will update the example to be more clear, thanks.

wurst/network/Network.wurst
SamuelMoriarty: Network performance degrades significantly with longer gamecache keys, since keys (as well as gamecache names themselves) also get sent over the network. This has been demonstrated by TriggerHappy previously in his findings for the Sync library.

https://www.hiveworkshop.com/pastebin/3a4f7861cb884e21312168bd654330585801/

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+		We also need to know the size of data being sent prior to starting the transmission,
+		so we have to store all of it in an intermediate buffer, which is DataStream.
+		Prior to sending, all strings in the DataStream are 'exploded' into a stream
```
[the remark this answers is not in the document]
SamuelMoriarty: Thanks!

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+		of integers, because SyncStoredString doesn't work.
```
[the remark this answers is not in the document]
SamuelMoriarty: Very old news. Source is TH's Sync library.

I tried this myself hoping it'd work on the newer patches somehow, but it simply doesn't sync strings correctly.
Might test this more in the future, but I have no reason to assume that it works.

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+		After sending, they are 'compacted' back into strings and written to the DataStream.
```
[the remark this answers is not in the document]
SamuelMoriarty: Thanks!

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+
+		GamecacheStream is a gamecache-based container with stream semantics for writing
+		integers, reals and booleans. There is a GamecacheStream for each primitive type,
+		integer, boolean, real and stringInts.
+
+		Network is the main class that coordinates DataStream and GamecacheStream and does
+		all the heavy lifting.
+
+		Before starting the transmission, the DataStream is locked into a non-writable, non-readable
+		mode to prevent anyone from trying to mess with it while the transmission is going.
```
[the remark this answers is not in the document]
SamuelMoriarty: They will receive an error with safety checks enabled.

wurst/network/Network.wurst
SamuelMoriarty: Would introduce useless overhead (key string stored in typecasting hashtable and translated back/forth to be used in another table) for slightly nicer syntax. Not worth it in a library imo.

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+			gcKeysCount++
+	else
+		for i = 0 to GC_KEYS.length() - 1
+			execute(() -> precomputeKeys(prefix + GC_KEYS.charAt(i), depth - 1))
+
+function precomputeKeys(integer length)
+	precomputeKeys("", length - 1)
+
+function precomputeLookup()
+	alphabetLookup = new Table()
```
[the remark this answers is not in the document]
SamuelMoriarty: The function is meant to be only called from within the init block, inside an execute() closure.

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+
+	for i = 0 to STRING_ALPHABET.length() - 1
+		let c = STRING_ALPHABET.charAt(i)
+
+		if (c.toUpperCase() == c and c.toLowerCase() != c) or c == "/"
+			alphabetLookup.saveInt(("00" + c).getHash(), i)
+		else
+			alphabetLookup.saveInt(c.getHash(), i)
+
+function testCollisions()
```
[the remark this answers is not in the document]
SamuelMoriarty: Thanks. Will do.

wurst/network/Network.wurst
SamuelMoriarty: The choice is between 1 Table and 4 HashMaps. HashMaps: Reals cannot be cast to ints safely without losing precision. Strings require an additional store in another table. I think a single Table here is justified.

Would totally use HashMaps if not for the above issue.

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+	// max characters to store per compacting round
+	private static constant integer CHARS_PER_COMPACT = 512
+
+	private Table buffer = new Table()
+	// stores strings as int-sequences
+	private Table stringBuffer = new Table()
+
+	private StreamMode currentMode
+
+	private integer integerWriteIndex = -1
```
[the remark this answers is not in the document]
SamuelMoriarty: Thanks

wurst/network/Network.wurst
SamuelMoriarty: Thanks

wurst/network/Network.wurst
SamuelMoriarty: Thanks

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+		destroy buffer
+
+	private function checkMode(StreamMode toCheck)
+		if currentMode != toCheck
+			if currentMode == StreamMode.WRITING
+				error("DataStream: trying to read from a stream that is currently writing")
+			else
+				error("DataStream: trying to write to a stream that is currently reading")
+
+	function writeInteger(integer value)
```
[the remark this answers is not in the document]
SamuelMoriarty: Whoops. My bad. A fluke on my part, it seems.

Doesn't Wurst use tabs primarily, though, and not spaces?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+
+				// prevent OP limit overrun by running this in separate thread
+				execute(() -> begin
+					explodeIndex++
+					stringBuffer.saveInt(explodeIndex, value.length())
+					// wrap loop counter in a box to allow modifying it from inside the closure
+					var j = new Box<integer>(0)
+					while j.value < value.length()
+						// split writes into sequences of no longer than CHARS_PER_COMPACT
+						execute(() -> begin
```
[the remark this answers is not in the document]
SamuelMoriarty: I agree, thanks, will do.

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+	function getRealCount() returns integer
+		return realWriteIndex + 1
+		
+	function getBooleanCount() returns integer
+		return booleanWriteIndex + 1
+		
+	function getStringIntCount() returns integer
+		return compactIndex + 1
+
+interface SynchronizerFunction
```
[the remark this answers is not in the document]
SamuelMoriarty: Internal functional interface for Network.onFinish() to take lambdas

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+	private GamecacheStream gcStringStream 
+
+	// the player that we are receiving data from
+	private player sender
+
+	// metadata about the transmission
+	private integer integerCount = 0
+	private integer realCount    = 0
+	private integer booleanCount = 0
+	private integer stringIntCount  = 0
```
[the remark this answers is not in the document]
SamuelMoriarty: sorry :v

wurst/network/Network.wurst
SamuelMoriarty: Thanks

wurst/network/Network.wurst
SamuelMoriarty: Thanks

wurst/network/Network.wurst
SamuelMoriarty: killmeplease

wurst/network/Network.wurst
```diff
@@ -0,0 +1,922 @@
...
+
+			// send all data
+			gcIntegerStream.sync()
+			gcBooleanStream.sync()
+			gcRealStream.sync()
+			gcStringStream.sync()
+		
+		// sync to receive data
+		synchronizer.sync()
+		synchronizer.onSynced(() -> receiveRound())
```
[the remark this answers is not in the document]
SamuelMoriarty: Thanks.

A Future or a Promise sounds like a good idea. Might explore it in the future (ha-ha)

I should probably make it more clear that the callback is supposed to be set in the same frame as .sync() is called. SimpleSynchronizer takes at least one round-trip time to complete, however, so here order doesn't matter. Calling `sleep(5)` would definitely break it, though.

wurst/network/Network.wurst
SamuelMoriarty: Thanks

wurst/network/SyncSimple.wurst
SamuelMoriarty: Thanks. Will do.

wurst/network/SyncSimple.wurst
SamuelMoriarty: 2^12 - 1 is `1111 1111 1111` in binary, one bit for each player. neutrals don't count.

wurst/network/Network.wurst
Cokemonkey11: HashMap provides compiletime safety - safety is never useless overhead.

But that doesn't mean it's the correct choice 100% of the time...

wurst/network/SyncSimple.wurst
Cokemonkey11: Maybe "Call `.sync()` for all game clients, after having called the Sync natives."

wurst/closures/Execute.wurst
Cokemonkey11: May I suggest `_currentCallback`? @Frotty ?

wurst/network/Network.wurst
Cokemonkey11: May I suggest

```
// Max amount to be written/read/synced per an execution, to avoid hitting the op-limit.
// Beware of changes in maximum value as compiler settings are adjusted.
```

wurst/network/Network.wurst
Frotty: You only use save/load/has int. You shouldn't need typecasting at all and thus there will be no overhead?!

wurst/network/Network.wurst
Cokemonkey11: May I suggest making all the documents clear on this unusual behavior?

"In multiplayer games, this package is for synchronizing data between game clients. It's useful for when one player is the source of game data, such as from gamecache or file IO."

wurst/network/Network.wurst
Frotty: in this case it's ok

wurst/network/Network.wurst
Cokemonkey11: Sorry, the purpose of my comment was to indicate that your code comment wasn't clear enough.

Rule of thumb: code comments explain why, not what.

wurst/closures/Execute.wurst
Frotty: why underscore in front? the other libs (closureforgroups) use `currentCallback` which should be fine

wurst/network/Network.wurst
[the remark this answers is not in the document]
Frotty: yes, the stdlib's packages are broken as well with 1.28.3+ or so.

wurst/network/Network.wurst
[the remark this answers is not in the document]
Frotty: Yes we mostly use tabs. But it should be fine as long as it's not mixed.

wurst/network/Network.wurst
Cokemonkey11: Stream: why does that make it a misnomer? are you imagining that UDP streams offer no guarantee but datagram ones do, or are you saying that the presence of a sequence of frames does not fit the definition of stream?

In both cases I disagree. Stream is pretty abstract. See https://tokio.rs/docs/getting-started/streams-and-sinks/

About safety checks: consider rephrasing the sentence so "MUST NOT" doesn't exist.

1) The API should do its best to not let someone do something wrong
2) MUST NOT is an angry warning that will, on its own, fall upon deaf ears.

wurst/network/Network.wurst
Cokemonkey11: "datagram" you mean when full set (stream) of datagrams have all been received?

This needs serious clarification imo.

"4. Register a callback function to be run on all game clients when the Network sync is complete."

wurst/network/Network.wurst
Cokemonkey11: I'm wondering if this warning is necessary. Will wurst compiler prevent you from use-after-free? @Frotty

wurst/network/Network.wurst
Cokemonkey11: Is this documented in JassDoc? @lep ?

Code needs a direct reference if you're going to both (1) make claims about perf, and (2) they're actually relevant

wurst/network/Network.wurst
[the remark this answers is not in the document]
Cokemonkey11: Code comment needs a source more than I do. Is this in jassdoc? @lep

wurst/network/Network.wurst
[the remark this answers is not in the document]
Cokemonkey11: "the DataStream is locked into an immutable state, in which incorrect mutation attempts will print warnings, so long as safety checks are not disabled."

wurst/network/Network.wurst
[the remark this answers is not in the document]
Cokemonkey11: Can you make some indication of this? I can suggest either

`// Only called in the init block.`, or

`function _onetimePrecomputeLookup()`

wurst/network/Network.wurst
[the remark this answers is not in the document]
Cokemonkey11: A previous version of the manual demanded tabs, but that's been relaxed and possibly even flip-flopped.

wurst/network/Network.wurst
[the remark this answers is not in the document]
Cokemonkey11: (Needs a comment)

wurst/network/Network.wurst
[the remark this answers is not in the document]
Cokemonkey11: I would urge you to either

* force onSynced to be called first, or
* at least always call it first in docs

Which helps indicate that you can't have TSA/context change between the two calls.

wurst/network/SyncSimple.wurst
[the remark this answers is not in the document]
Cokemonkey11: It would be great to have some amount of this explanation instead of just "This library can be used to 'synchronize' players after some async network action has been performed" which sounds like "I did some magical desync thing which I can't debug and now I'm using Network to fix my problems".

wurst/network/SyncSimple.wurst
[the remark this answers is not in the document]
Cokemonkey11: Needs documentation and source - is it in JassDoc? @lep ?

wurst/network/SyncSimple.wurst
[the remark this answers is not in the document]
Cokemonkey11: > This is due to the fact that the WC3 game protocol is built on top of TCP,

What's the actual value of explaining why [mathematical deterministic thing] works in [mathematically deterministic way]?

This is good background knowledge but not so much documentation.

wurst/network/SyncSimple.wurst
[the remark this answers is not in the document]
Cokemonkey11: By Concrete Demo I mean actual testmap. Here's an example of PR where testmap was warranted: https://github.com/wurstscript/WurstScript/pull/472

wurst/network/SyncSimple.wurst
Cokemonkey11: Needs a comment. Also would be nice if wurst had `0b1111_1111_1111` notation for ints @Frotty

wurst/closures/Execute.wurst
Cokemonkey11: That's fine too. Why underscore? Costs very little and benefit is slightly improved indication of "this is not public, and its use is very specific/controlled by this package"

wurst/network/Network.wurst
SamuelMoriarty: Not true, I also store strings and reals?

SamuelMoriarty:
I've updated the PR, implementing most of the requests outlined above.
I've also done some refactoring in Network. Here's a short summary (off the top of my head)
* Added Force API
* Moved Buffer classes into wurst/data, added different implementations like HashBuffer and StringBuffer (to be used later for file I/O), also added unit tests for them
* Split apart Network into multiple sub-packages: GamecacheBuffer, GamecacheKeys, Metadata, Network and StringEncoder
* Metadata functionality has been externalized
* String-encoding has been externalized
* Network now sends first payload together with metadata, to avoid one redundant round-trip

wurst/data/Buffer.wurst
```diff
@@ -0,0 +1,602 @@
...
+	private function getNextChunkSize() returns integer
+		return chunks.getFirst().content.length()
+
+	// loads the next chunk into the buffer
+	private function nextChunk()
+		if chunks.size() == 0
+			fail(BufferFailureMode.EOF, "reached EOF")
+			return
+
+		let chunk = chunks.dequeue()
```
Frotty: using push and then dequeue kinda doesn't make sense. Please stick to sensible naming conventions of the intended datatype.

wurst/data/Buffer.wurst
```diff
@@ -0,0 +1,602 @@
...
+		checkWrite()
+		pushTypeIdentifier(ValueType.STRING)
+		pushString(value.length().toString())
+		pushTerminator()
+		pushString(value)
+
+	override function writeBoolean(bool value)
+		checkWrite()
+		pushTypeIdentifier(ValueType.BOOLEAN)
+		if value
```
Frotty: value.toString()

wurst/data/Buffer.wurst
```diff
@@ -0,0 +1,602 @@
...
+			pushString("1")
+		else
+			pushString("0")
+
+	private function readIntInternal() returns int
+		return S2I(popUntilTerminator())
```
Frotty: .toInt(), avoid using natives directly

wurst/data/Buffer.wurst
```diff
@@ -0,0 +1,602 @@
...
+
+	override function readInt() returns int
```
Frotty: Either make read-functions inline friendly or use ternary op to shorten

wurst/data/Buffer.wurst
```diff
@@ -0,0 +1,602 @@
...
+				fail(BufferFailureMode.MALFORMED_INPUT, "tried to read string of length " + length.toString() +
+					", but max is " + MAX_BUFFER_SIZE.toString())
+				return null
+			return popString(length)
+		return null
+
+	override function readBoolean() returns bool
+		checkRead()
+		if checkType(ValueType.BOOLEAN)
+			let value = popString(1)
```
Frotty: can be linlined

wurst/data/BufferTests.wurst
```diff
@@ -0,0 +1,129 @@
...
+
+	buffer.flush()
+
+	for i = 0 to 255
+		buffer.readInt().assertEquals(i)
+		buffer.readReal().assertEquals(i + 0.125)
+		buffer.readBoolean().assertEquals(i div 2 == 0)
+		buffer.readString().assertEquals(i.toString())
+
+// disabled while there are some bugs preventing it from working correctly (related to the compiler)
```
Frotty: what's the status on this? should be fixed?

wurst/network/StringEncoder.wurst
```diff
@@ -0,0 +1,130 @@
...
+
+init
+	execute(() -> precomputeLookup())
+	execute(() -> testCollisions())
+/**
+	Utility class to convert a string to/from a series of ints corellating to positions in STRING_ALPHABET.
+**/
+public class StringEncoder
+	// max characters to store per packing round
+	private static constant int CHARS_PER_COMPACT = 512
```
Frotty: omit types

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,177 @@
...
+		end)
+
+		registerPlayerEvent(EVENT_PLAYER_LEAVE, () -> begin
+			let who = GetTriggerPlayer()
+
+			for synchronizer in SimpleSynchronizer
+				synchronizer.onPlayerLeave(who)
+		end)
+		
+init
```
Frotty: just put the registerEvent calls into init block?

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,177 @@
...
+		..addAbility(DUMMY_GHOST)
+		..pause()
+		..setScale(0)
+		reverseLookup[dummy.getIndex()] = this
+
+	ondestroy
+		reverseLookup[dummy.getIndex()] = null
+		dummy.remove()
+
+	/**
```
Frotty: these should be 1-line hotdocs

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,177 @@
...
+	
+public interface SynchronizationCallback
+	function onSynchronized()
+
+public class SimpleSynchronizer
+	use LinkedListModule
+
+	private static constant DUMMY_PLAYER = players[PLAYER_NEUTRAL_PASSIVE]
+	private static constant DUMMY_ID     = 'hfoo'
+	private static constant DUMMY_GHOST  = 'Aeth'
```
Frotty: Already exists in Basics.wurst as GHOST_VIS_ID

wurst/network/GamecacheKeys.wurst
```diff
@@ -0,0 +1,33 @@
...
+
+// characters that can be used as GC keys
+// their amount also determines the size of a single 'round'
+constant string GC_KEYS = " !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`{|}~"
+
+public class GamecacheKeys
+	static string array keys
+	static var count = 0
+
+	static function getKey(int i) returns string
```
Frotty: if the class is already name **Keys this should just be "get"

wurst/network/Metadata.wurst
```diff
@@ -0,0 +1,65 @@
...
+	return metadata(this.intCount, this.realCount, value, this.asciiIntCount, this.syncCount)
+
+public function metadata.setAsciiIntCount(int value) returns metadata
+	return metadata(this.intCount, this.realCount, this.booleanCount, value, this.syncCount)
+
+public function metadata.setSyncCount(int value) returns metadata
+	return metadata(this.intCount, this.realCount, this.booleanCount, this.asciiIntCount, value)
+
+/**
+	Simple wrapper class for metadata.
```
Frotty: What exactly is metadata?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,506 @@
+package Network
+
+import SyncSimple
+import ErrorHandling
+import Execute
+import Buffer
+
+import initlater Metadata
```
Frotty: You should avoid initlater if possible. I think for the most parts you just access constants from another package.
I would say this means either move those into a config package or move the init to the other package

wurst/network/Network.wurst
```diff
@@ -0,0 +1,506 @@
...
+import initlater GamecacheKeys
+import initlater GamecacheBuffer
+import initlater StringEncoder
+
+/**
```
Frotty: me personally, I prefer more compactz hotdoc with no empty lines. If u rpefer this you can leave it however.

wurst/network/Network.wurst
```diff
@@ -0,0 +1,506 @@
...
+	private var stringEncoder = new StringEncoder
+
+	// gamecache buffers to sync data in rounds
+	private var gcIntBuffer     = new GamecacheBuffer(DATA_CACHE, mkey, BufferType.INTEGER)
+	private var gcRealBuffer    = new GamecacheBuffer(DATA_CACHE, mkey, BufferType.REAL)
+	private var gcBooleanBuffer = new GamecacheBuffer(DATA_CACHE, mkey, BufferType.BOOLEAN)
+	private var gcStringBuffer  = new GamecacheBuffer(STRING_CACHE, mkey, BufferType.INTEGER)
+
+	// metadata container
+	private var metadataStore = new MetadataStore
```
Frotty: ()

wurst/network/Network.wurst
```diff
@@ -0,0 +1,506 @@
...
+
+	// the player that we are receiving data from
+	private player sender
+
+	// metadata about the transmission
+	private var meta = metadata(0, 0, 0, 0, 0)
```
Frotty: make some EMPTY_META constant for the null element

wurst/network/Network.wurst
```diff
@@ -0,0 +1,506 @@
...
+	private var metaReceived = false
+
+	// these double as write counters for the sender and read counters for the receivers
```
Frotty: why are these not a tuple as well?

wurst/network/Network.wurst
```diff
@@ -0,0 +1,506 @@
...
+		destroy gcBooleanBuffer
+		destroy gcStringBuffer
+		destroy stringEncoder
+		destroy dataBuffer
+		destroy metadataStore
+
+	function getData() returns HashBuffer
+		return dataBuffer
+
+	// sends info about the amount of data to be expected to each player
```
Frotty: hotdoc

wurst/network/Network.wurst
```diff
@@ -0,0 +1,506 @@
...
+
+		if localPlayer == sender
+			// push all strings into the string encoder
+			execute(() -> begin
+				while dataBuffer.hasString()
+					stringEncoder.pushString(dataBuffer.readString())
+			end)
+
+			// encode all strings into ints
+			execute(() -> begin
```
Frotty: doesn't need begin/end

wurst/network/Network.wurst
```diff
@@ -0,0 +1,506 @@
...
+		dataBuffer.setMode(BufferMode.LOCKED)
+
+		// get the amount of individual entities to be synced for each type
+		let intCount      = dataBuffer.getIntCount()
+		let realCount     = dataBuffer.getRealCount()
+		let booleanCount  = dataBuffer.getBooleanCount()
+		let asciiIntCount = stringEncoder.getIntCount()
+		
+		// calculate max length in each entity buffer
+		let maxLength = max(max(intCount, realCount), max(booleanCount, asciiIntCount))
```
Frotty: should probably make a max(x,y,z) for maths package

wurst/network/Network.wurst
```diff
@@ -0,0 +1,506 @@
...
+				end)
+
+			dataBuffer.setMode(BufferMode.READ)
+			dataBuffer.resetRead()
+			currentState = NetworkState.FINISHED
+			finishCallback.onFinish(dataBuffer)
+			destroy this
+
+	// this is the function to start sending all data in the intermediate dataBuffer buffer
+	function start()
```
Frotty: can combine start and onFinish I think

wurst/network/SyncSimple.wurst
```diff
@@ -0,0 +1,177 @@
...
+		There may be other usages related to async network events as well.
+
+	Usage:
+		1. Create a new SimpleSynchronizer using:
+			let synchronizer = new SimpleSynchronizer()
+
+		2. Send some network events on one (or more) player using the Sync natives (or something else)
+			if sender == localPlayer
+				// value1 and value2 are some local values only known to localPlayer
+				SyncStoredInteger(gc, mkey, vkey1, value1)
```
Frotty: use extension funcs

wurst/data/Buffer.wurst
SamuelMoriarty: should I use .add()/.dequeue() pair instead?

wurst/_handles/Trigger.wurst
```diff
@@ -1,5 +1,6 @@
...
```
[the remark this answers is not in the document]
SamuelMoriarty: Done

wurst/data/Buffer.wurst
```diff
@@ -0,0 +1,602 @@
...
+		4. "s10|0123456789" - string 0123456789
+		5. "s5|01234i10|b1b0r1.0005|" - string 012345, integer 10, bool true, bool false, real 1.0005
+**/
+public class StringBuffer extends AbstractBuffer
+	private static constant TERMINATOR = "|"
+	private static constant MAX_BACKBUFFER_SIZE = 512
+	private static constant MAX_BUFFER_SIZE = 1024
+
+	// list of "chunks" that can grow/shrink as required
+	private var chunks = new LinkedList<ChunkElement>()
```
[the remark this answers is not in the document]
SamuelMoriarty: Done

wurst/data/Buffer.wurst
```diff
@@ -0,0 +1,602 @@
...
+	private function checkDataAvailable()
+		if not isDataAvailable()
+			fail(BufferFailureMode.EOF, "reached EOF")
+
+	private function getNextChunkSize() returns integer
+		return chunks.getFirst().content.length()
+
+	// loads the next chunk into the buffer
+	private function nextChunk()
+		if chunks.size() == 0
```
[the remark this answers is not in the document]
SamuelMoriarty: Done

wurst/data/Buffer.wurst
SamuelMoriarty: Done

wurst/data/Buffer.wurst
SamuelMoriarty: Done

wurst/data/Buffer.wurst
SamuelMoriarty: Done

wurst/data/Buffer.wurst
SamuelMoriarty: Done

wurst/data/BufferTests.wurst
SamuelMoriarty: Works correctly now, re-enabled test

wurst/network/GamecacheKeys.wurst
SamuelMoriarty: Done

wurst/network/Metadata.wurst
SamuelMoriarty: Updated hotdoc to reflect class purpose better

wurst/network/Network.wurst
SamuelMoriarty: Moved config out into separate package

wurst/network/Network.wurst
SamuelMoriarty: Done

wurst/network/Network.wurst
SamuelMoriarty: Reused metadata tuple for counters

wurst/network/Network.wurst
SamuelMoriarty: Done

wurst/network/Network.wurst
SamuelMoriarty: Done

wurst/network/Network.wurst
SamuelMoriarty: Added min/max variants for up to 6 arguments in Maths package

wurst/network/Network.wurst
SamuelMoriarty: Done

wurst/network/StringEncoder.wurst
SamuelMoriarty: Done

wurst/network/SyncSimple.wurst
SamuelMoriarty: Done

wurst/network/SyncSimple.wurst
SamuelMoriarty: Done

Frotty:
Sweet 🌴

PR #20, opened 2017-12-08
remove compiletime call in OnUnitEnterLeave

IgorSamurovic:
Apparently, ever since the order of compiletime calls has become deterministic, the compiletime(x) function is just messing things up. I had to remove this in order to get the same order in compiletime/game.

### DIFF
--- a/wurst/event/OnUnitEnterLeave.wurst
+++ b/wurst/event/OnUnitEnterLeave.wurst
@@ -49,9 +49,7 @@ init
 		players[i].setAbilityAvailable(ABILITY_ID, false)
 
 	// Create the enter event
-	CreateTrigger()..registerEnterRegion(boundRegion, Filter(() -> begin
-		prepareUnit(GetFilterUnit())
-	end))
+	CreateTrigger()..registerEnterRegion(boundRegion, Filter(() -> prepareUnit(GetFilterUnit())))
 
 	// Create the leave event
 	registerPlayerUnitEvent(EVENT_PLAYER_UNIT_ISSUED_ORDER, () -> begin
--- a/wurst/event/OnUnitEnterLeave.wurst
+++ b/wurst/event/OnUnitEnterLeave.wurst
@@ -10,7 +10,6 @@ import AbilityObjEditing
 import RegisterEvents
 import ObjectIdGenerator
 import MapBounds
-
 import ClosureTimers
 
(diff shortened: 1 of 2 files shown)


Frotty:
I don't think so. Compl-expressions are the way to go to have ids _only_ generated at compiletime. 
You have to make sure to not use it outside of those.

IgorSamurovic:
Alright, I found the actual issue that was plaguing the map, it works for preplaced units on a brand new map now, too. Previously it wouldn't trigger onEnter on units preplaced by the editor.

wurst/event/OnUnitEnterLeave.wurst
```diff
@@ -11,18 +11,20 @@ import RegisterEvents
 import ObjectIdGenerator
 import MapBounds
```
Frotty: prune empty line

wurst/event/OnUnitEnterLeave.wurst
```diff
@@ -11,18 +11,20 @@ import RegisterEvents
...
 
+import ClosureTimers
+
 /* 	Provides event API for units entering and leaving the map.
 	The event will also fire for preplaced units on the map. */
 
-constant eventTrigger = CreateTrigger()
-unit tempUnit = null
+let eventTrigger = CreateTrigger()
```
Frotty: why change to let? I prefert `constant` outside of functions tbh

IgorSamurovic:
Alright, so I did a lot of tinkering about this, and I noticed a problem -

No matter how you use the timer for preplaced units, it's impossible to know whether you'll catch them in the group loop or via an enter event because some scripts will be initialized after OnUnitEnterLeave, some will be initialized before it, which means that, while ensuring that the unit will get processed at all costs (I moved the enum inside the nulltimer), there's a possibility of the unit triggering the enter event.

So, in order to remove the possibility of double processing the unit, I put everything in the null timer. After the null timer ticks:
1) units are first going to be enum'd across the map, and processed.
2) Unit entering is initialized
3) Unit leaving is initialized

This means that no matter where you use onEnter, as long as it is on map init, it will be processed for all units after nullTimer ticks.

Also, DamageDetection didn't work for non-preplaced units, because it used the old onEnter semantics which used Filtered unit for some reason. Now it simply adds an onEnter action, and works for both preplaced and units added after init and after the null timer.

PR #25, opened 2018-01-02
add GroupUtils and change Group.wurst to use its functionalities

IgorSamurovic:
Also, .destr() now calls .release()
The code is tested in game, and it works with my systems (that use a bunch of groups) without leaking. I can't guarantee more than that. This is impossible to test because get handle id is unavailable during compiletime. This also breaks GroupTests.wurst because it now uses getGroup() during group iteration.

### DIFF
--- a/wurst/_handles/Group.wurst
+++ b/wurst/_handles/Group.wurst
@@ -82,4 +82,4 @@ public function group.next() returns unit
 	return iterUnit
 	
 public function group.close()
-	this.destr()
+	this.release()
--- a/wurst/_handles/Group.wurst
+++ b/wurst/_handles/Group.wurst
@@ -68,18 +68,22 @@ public function group.getRandomUnit() returns unit
 
 group iterGroup
 
+/** Creates a new iterator for this group. */
 public function group.iterator() returns group
 	iterGroup = getGroup()
 	ForGroup(this, () -> iterGroup.addUnit(GetEnumUnit()))
 	return iterGroup
 
+/** Returns whether the iterator has the next item */
 public function group.hasNext() returns boolean
 	return FirstOfGroup(this) != null
-	
+
+/** Returns the next item from the iterator */
 public function group.next() returns unit
 	let iterUnit = FirstOfGroup(this)
 	GroupRemoveUnit(this, iterUnit) 
 	return iterUnit
 	
+/** Closes the iterator, releaseing the group */
 public function group.close()
 	this.release()
--- a/wurst/_handles/Group.wurst
+++ b/wurst/_handles/Group.wurst
@@ -33,7 +33,7 @@ public function group.removeUnit(unit u)
 	GroupRemoveUnit(this, u)
 	
 public function group.destr()
-	this.release()
+	DestroyGroup(this)
 	
 public function group.contains(unit u) returns boolean
 	return IsUnitInGroup(u, this)
@@ -66,7 +66,6 @@ public function group.getRandomUnit() returns unit
 	
 /* Group iterator */
 
-unit iterUnit
 group iterGroup
 
 public function group.iterator() returns group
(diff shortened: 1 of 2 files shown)


wurst/util/GroupUtils.wurst
```diff
@@ -0,0 +1,134 @@
...
+let used = new HashMap<group, bool>
+group array stack
+var numStack = 0
+var numTotal = 0
+
+var shownMaxError = false
+
+function createGroups(int number)
+    let maxCreatePerCycle = 256
+    let actualLimit = IMinBJ(8191, GROUP_NUMBER_LIMIT)
```
Frotty: Use max and min from Maths package.

wurst/util/GroupUtils.wurst
```diff
@@ -0,0 +1,134 @@
...
+    var toCreate  = IMaxBJ(0, numTarget - overflow)
+
+    while toCreate > 0
+        let createNow = IMinBJ(toCreate, maxCreatePerCycle)
+        toCreate -= createNow
+
+        if compiletime 
+            numTotal += createNow
+            for i = 1 to createNow
+                let g = CreateGroup()
```
Frotty: inline this

wurst/util/GroupUtils.wurst
```diff
@@ -0,0 +1,134 @@
...
+    let actualLimit = IMinBJ(8191, GROUP_NUMBER_LIMIT)
+    let numTarget = numTotal + number 
+    let overflow  = IMaxBJ(0, numTarget - actualLimit)
+    var toCreate  = IMaxBJ(0, numTarget - overflow)
+
+    while toCreate > 0
+        let createNow = IMinBJ(toCreate, maxCreatePerCycle)
+        toCreate -= createNow
+
+        if compiletime
```
Frotty: one should just implement ForForce for compiletime instead of this conditional imo.

wurst/util/GroupUtils.wurst
```diff
@@ -0,0 +1,134 @@
...
+        Log.warn("Attemping to release ENUM_GROUP!")
+    else
+        if this.isRecyclable()
+            if this.isUsed()
+                this.recycle()
+            else
+                Log.warn("Attemping to release an already released group!")
+        else
+            this.clear()
+            DestroyGroup(this)
```
Frotty: .destr()

wurst/_handles/Group.wurst
```diff
@@ -65,20 +64,23 @@ public function group.getRandomUnit() returns unit
...
-	bj_groupAddGroupDest = CreateGroup()
-	ForGroup(this, function GroupAddGroupEnum)
-	return bj_groupAddGroupDest
+	iterGroup = getGroup()
+	ForGroup(this, () -> iterGroup.addUnit(GetEnumUnit()))
+	return iterGroup
 
 public function group.hasNext() returns boolean
-	return FirstOfGroup(this) != null
+	iterUnit = FirstOfGroup(this)
```
Frotty: this will break if .next() isn't called instantly after hasNext or if someone only calls .next()
```
while itr1.hasNext()
    if itr2.hastNext()
        let u1 = itr1.next()
        let u2 = itr2.next()  // u1 and u2 will be equal

OR


someGroup.next() // ignore one unit
let second = someGroup.next() // this one will still be the first one

```
I know these use cases shouldn't happe and the user should call .next() only, but instantly after .hasNext() anyways, but yeah :/

wurst/util/GroupUtils.wurst
```diff
@@ -0,0 +1,134 @@
...
+    return retVal
+
+function group.recycle()
+    this.clear()
+    push(this)
+
+/** Returns a reusable group from the GroupUtils stack is possible.
+    Creates new groups if none are found on the reusable stack.
+    If that is not possible, returns a non-recyclable group and displays a warning. */
+public function getGroup() returns group
```
Frotty: move API functions above internal functions

IgorSamurovic: API moved to the top

Frotty:
Thanks !

PR #26, opened 2018-01-03
Network, Execute fixes & FileIO

SamuelMoriarty:


wurst/file/IOTaskExecutor.wurst
```diff
@@ -0,0 +1,149 @@
...
+	private var lastExecution = 0.0
+
+	construct(real delay)
+		this.delay = delay
+			
+	protected static function updateInstances()
+		for executor in TimedIOTaskExecutor
+			executor.update()
+
+	// called when an executor instance has started working
```
Frotty: convert to hot doc, also the ones below

wurst/file/IOTaskExecutor.wurst
```diff
@@ -0,0 +1,149 @@
...
+			//updater.pause()
+
+	// updates this instance if necessary
+	private function update()
+		if not finished and lastExecution + delay <= currentTime
+			executeNext()
+			lastExecution = currentTime
+			if finished
+				runningCount--
+				// stop the timer if necessary
```
Frotty: "if necessary" ?

wurst/file/Persistable.wurst
```diff
@@ -0,0 +1,165 @@
...
+	data in a safe, synchronous manner 
+**/
+public abstract class Persistable implements BufferSerializable
+	private PersistentStore store
+	/** The owner of the object, from whom and to who this object will be saved. **/
+	protected player owner
+
+	construct(player owner)
+		this.owner = owner
+		store = new PersistentStore(this, owner)
```
Frotty: is the store ever destroyed?

wurst/file/Persistable.wurst
```diff
@@ -0,0 +1,165 @@
...
+		return 0
+	/** 
+		Internal function that returns the path that the object should be saved under.
+		You can override this if you feel the need to change the paths under which
+		your objects are saved.
+	**/
+	protected function getPath() returns string
+		return PERSISTABLE_MAP_PREFIX + "/" + getClassId() + "/" + getId()
+
+	/* You must implement these */
```
Frotty: then they should probably be at the top :P

wurst/file/Persistable.wurst
```diff
@@ -0,0 +1,165 @@
...
+		if localPlayer == owner
+			reader.load(() -> begin
+				var buffer = new StringBuffer(MAX_PACKET_LENGTH)
+				// a boolean for keeping track of whether the load was succesful or not
+				network.getData().writeBoolean(true)
+				// read the data from the file into the buffer, and then transfer it into the network's buffer
+				buffer.populateFromFile(reader)
+				buffer.transfer(network.getData())
+				// if there was an error during transfer, we scrap the data...
+				if not buffer.isGood()
```
Frotty: "isGood" should be replaced by something more descriptive

wurst/file/IOTaskExecutor.wurst
```diff
@@ -0,0 +1,149 @@
...
+		if not finished
+			error("TimedIOTaskExecutor: trying to start an executor that is already running")
+
+		finished = false
+		runningCount++
+		// make sure the updater timer is running
+		onStart()
+
+init
+	TimedIOTaskExecutor.updater.startPeriodic(ANIMATION_PERIOD, function TimedIOTaskExecutor.updateInstances)
```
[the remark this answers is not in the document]
SamuelMoriarty: Tasks are sometimes run locally only for one player. Starting/stopping a timer locally will most likely cause a desync (at least in my memory it does).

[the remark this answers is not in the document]
Frotty: start/stop it non-locally ❓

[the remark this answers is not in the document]
SamuelMoriarty: We can't do that since tasks are created locally for each player.

wurst/closures/Execute.wurst
```diff
@@ -1,37 +1,106 @@
...
 public function execute(ForForceCallback c)
 	pushCallback(c)
 	executeForce.forEach(function executeCurrentCallback)
+	let success = getCurrentCallbackSuccess()
+	if not success
+		error("execute: callback has failed")
 	popCallback()
+
+interface LimitedExecuteCondition
+	function check() returns boolean
```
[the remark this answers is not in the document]
SamuelMoriarty: The 'iterations' here is a completely internal variable that is used to reset the OP limit when iteration count is exhausted. This value can't be used to control the outer loop, because it gets reset to 0 each `iterations` times.

Without a `condition`, this function will simply loop forever. The condition is used to break the loop.

wurst/closures/Execute.wurst
```diff
@@ -1,37 +1,106 @@
...
+
+		// restart the execute if necessary
+		if condition.check()
+			limitedExecuteInternal(iterations, condition, action)
+	end)
+
+/**
+	This is an utility function which you can use to process arbitrarily large loops.
+	It functions like a while loop that will run action.run() while condition.check() == true,
+	however it will restart the execute after the specified amount of iterations.
```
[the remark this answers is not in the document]
SamuelMoriarty: When processing arbitrarily large loops you are always going to need some sort of 'cutoff' value that is chosen based on some heuristic (and testing) to reset the OP limit.

The iteration count is going to depend on the payload size inside `action`, so you can't go around fine-tuning it.

I wrote this function to avoid a lot of boilerplate code used to reset the OP limit in Network and other libraries of my own, and I originally handled iteration wholly inside client code, and it was a mess.

Contrary to that mess, `limitedExecute` (which I renamed to `executeWhile` now), acts simply as a while loop with a condition and an action, and a specifier when to restart the thread. That way, we don't incur too much overhead because of excessive `execute()` calls (which are pretty expensive, compared to regular func calls).

Also, absolute iterations aren't always known ahead of time, as is often the case with while loops - hence the need for the `condition` check.

wurst/file/IOTaskExecutor.wurst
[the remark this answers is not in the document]
SamuelMoriarty: Nevermind, it doesn't actually cause a desync.

wurst/_handles/Playercolor.wurst
```diff
@@ -0,0 +1,36 @@
...
+public function int.toPlayerColor() returns playercolor
+	return lookup[this]
+
+public function playercolor.toInt() returns int
+	var value = 0
+	for i = 0 to 15
+		if lookup[i] == this
+			value = i
+			break
+	return value
```
Cokemonkey11: missing eof lf

wurst/closures/Execute.wurst
```diff
@@ -1,37 +1,125 @@
...
+interface LimitedExecuteCondition 
+	function check() returns boolean 
+ 
+interface LimitedExecuteAction 
+	function run() 
+ 
+function executeWhileInternal(int resetCount, LimitedExecuteCondition condition, LimitedExecuteAction action) 
+	execute(() -> begin 
+		var i = 0 
+		// iterate as many times as needed
```
Cokemonkey11: comment not useful

wurst/data/Buffer.wurst
```diff
@@ -565,16 +579,33 @@ public class StringBuffer extends AbstractBuffer
 		checkRead()
 		return checkType(ValueType.BOOLEAN) ? (popString(1) == "1") : false
 
+	override function clear()
+		for chunk in chunks
+			destroy chunk
```
Cokemonkey11: `\t` ⁉️

wurst/file/MultiIO.wurst
```diff
@@ -0,0 +1,237 @@
...
+		4. Destroy the instance
+
+		Note: The class can be used again after saving once.
+**/
+public class FileWriter extends AbstractFile
+	/**
+		If multiMode is true, then path should be the path to a folder where the data will be stored.
+		Otherwise, the path should be a file path.
+
+		Delay is the number of seconds to wait between each file write operation. You should use
```
Cokemonkey11: `\t` ⁉️

wurst/file/Persistable.wurst
```diff
@@ -0,0 +1,176 @@
...
+import SimpleIO
+import BufferAdapters
+import public Buffer
+
+/**
+	This is the folder that all your map objects will be serialized in.
+	You should override this to prevent conflicts with other maps using this class.
+**/
+@configurable
+constant PERSISTABLE_MAP_PREFIX = "persistent"
```
Cokemonkey11: needs comment

wurst/file/SimpleIO.wurst
```diff
@@ -0,0 +1,139 @@
+package SimpleIO
+
+import ErrorHandling
+import StringUtils
+
+/** Maximum amount of packets per single readable file. */
+public constant PACKETS_PER_FILE = 16 	
+/** Maximum payload size of a single readable packet. */
+public constant MAX_PACKET_LENGTH = 209
```
Cokemonkey11: magic number?

Cokemonkey11:
Overall this is pretty brilliant. I think it's the biggest PR on stdlib. I hope Frotty has a backup of this code in case something happens.

As it stands, the code quaility is very high and I think you've balanced perfection and pragmatism well.

The nature of this stuff is fragile, though. There's necessarily a lot of non-generic type stuff here, bit twiddling, magic values, data reinterpretation, etc. I think that's fine.

My final opinion is that this could go two ways:

* stdlib isn't "stable", so this fragile code can get merged just fine without anyone worrying. After all, NASA is not writing wurst code. In this route, I think this MR needs some squashing/otherwise interactive rebasing.

* This huge, fairly specialised, fragile code probably could be in a separate lib - which is a great opportunity for us to play with dependency manager, and otherwise just makes sense.

I'll let you and Frotty decide which of these variants is the right one. In any case, *please someone get a backup*, and also thanks again, this is really, really cool 😎

Frotty:
I agree. I think it could be in stdlib to promote easy persistable data.
On the other hand, most players don't even have local files enabled, making this impractical for use in "any map".
If loading will ever be possible without local files, this would be killer stdlib feature tho imo

wurst/_wurst/ErrorHandling.wurst
```diff
@@ -8,10 +8,16 @@ import GameTimer
 import MagicFunctions
 import Hashtable
 
-constant MUTE_ERROR_DURATION = 60
+constant MUTE_ERROR_DURATION = 5
```
Frotty: I don't think this is a good idea. Should probably make the value @configurable though

Frotty:
thanks a lot 🎓

PR #27, opened 2018-01-04
add the ability to retrieve set ids from definitions

IgorSamurovic:
This is code that I've been using for a while but always put off adding it to the stdlib. My idea is to make data directly set by the constructor retrievable afterwards, this makes it easier to create object data frameworks since it allows you to pass newly constructed definitions as parameters, and then have your system decide what to do with the ids (and where and how to store them).


Frotty:
Thanks ❤️

PR #28, opened 2018-01-05
add euler's constant to the Maths package

IgorSamurovic:
I think more math functionalities should be added too, but will have to think about it.


Frotty:
You alignment is off. Also, shouldn't it just be "E"?

PR #31, opened 2018-01-07
add typecasting as a default import because let's be real everyone uses it all the time

IgorSamurovic:


Frotty:
Thank you :bowtie:

PR #33, opened 2018-01-08
Quest and Questitem extensions

cljnilsson:
Nuff said.

### DIFF
--- a/wurst/_handles/Quest.wurst
+++ b/wurst/_handles/Quest.wurst
@@ -1,5 +1,7 @@
 package Quest
 
+import QuestItem
+
 public function quest.setRequired(boolean req)
     QuestSetRequired(this, req)
 
@@ -34,7 +36,7 @@ public function quest.remove()
     DestroyQuest(this)
 
 
-    public enum QuestState
+public enum QuestState
     FAILED
     COMPLETED
     UNDISCOVERED
@@ -79,8 +81,8 @@ public class Quest
             case state.DISCOVERED
                 q.discovered(true)
 
-    function addObjective(string text) returns QuestObjective
-        return new QuestObjective(q, text)
+    function addObjective(string text) returns QuestItem
+        return new QuestItem(q, text)
         
     ondestroy
         q.remove()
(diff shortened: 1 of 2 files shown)


wurst/_handles/QuestItem.wurst
```diff
@@ -0,0 +1,24 @@
+package QuestItem
+
+public function questitem.description(string text)
```
Frotty: Shouldn't this be "setDescription" like the other ones?

wurst/_handles/Quest.wurst
```diff
@@ -0,0 +1,86 @@
...
+    return IsQuestCompleted(this)
+
+public function quest.flash()
+    FlashQuestDialogButton()
+
+public function quest.remove()
+    DestroyQuest(this)
+
+
+    public enum QuestState
```
Frotty: "public enum QuestState" should not be indented, this is causing the travis build to fail

wurst/_handles/Quest.wurst
```diff
@@ -0,0 +1,86 @@
...
+public function quest.discovered(boolean b)
+    QuestSetDiscovered(this, b)
+
+public function quest.isEnabled() returns boolean
+    return IsQuestEnabled(this)
+
+public function quest.isCompleted() returns boolean
+    return IsQuestCompleted(this)
+
+public function quest.flash()
```
Frotty: This shouldn't be an extension function, or?

cljnilsson:
I think it should be an extension unless I am missing something entirely. Everything else is fixed.

Frotty:
FlashQuestDialogButton doesn't take a quest handle though. And hence it doesn't change whatever quest you call it upon.

wurst/_handles/QuestItem.wurst
```diff
@@ -0,0 +1,24 @@
...
+    construct(quest q, string text)
+        parent = q
+        qi = QuestCreateItem(parent)
+        qi.setDescription(text)
+
+    function isCompleted() returns boolean
+        return qi.isCompleted()
+    
+    function complete(boolean b)
+        qi.complete(b)
```
Cokemonkey11: missing eof lf

cljnilsson:
Ohhh. Good point.

Frotty:
so will you push a fix?? 😆

Frotty:
I guess not. Well thanks anyway 👍

PR #37, opened 2018-01-10
add IterableMap to HashMap, update and test HashList 

IgorSamurovic:
also added copy methods as well as prototype constructors for all data types that allow it


Frotty:
Add HashList unit tests please 🔞

wurst/data/HashSet.wurst
```diff
@@ -0,0 +1,26 @@
+package HashSet
+import HashList
```
Frotty: should be import public

Frotty:
gud

PR #38, opened 2018-01-10
Add Soundsets class for units soundsets

karlek:
I extracted all unit soundsets so we can have static checking of soundsets in wurst 🌴 

__Note:__ I haven't tested all the soundsets

__Extraction of names__

`War3.txt` is the text file containing the paths of `War3.mpq`, and the same with `War3xlocal.txt`.

The command I ran to extract the names:
```shell
$ cat War3.txt War3xlocal.txt | grep '^Units.*\.wav' | awk -F '\\' '{print $(NF-1)}' | sort -u > soundsets
```

Cheers from
Karlek & Kranx


Frotty:
very nice 👍 thanks 🌴 
We should probably split up the Assets package into smaller ones and then have the Asset package be one with only public imports of the smaller packages.

karlek:
We noticed yesterday that Goblin Alchemist and Tinkerer are missing. I guess that this only concerns the neutral passive heroes that were added through patches. I'll make a new PR to add them. @Frotty can you think of any other units that have been added through patches?

Btw, I can make a PR for your proposal of splitting up the Assets class into separate files. Shall we still call it assets?

Jampi0n:
The Firelord was added later as well. You should also check their summons: Clockwork Goblin and Lava Spawn.
Not sure if Volcano or Pocket Factory have sound sets.
I doubt other units are affected, because tavern heroes are basically the only thing you can add in later patches.

Frotty:
Pretty much all units from the expansion eh?
I didn't check all names, I suppose you didn't take data from both mpqs?

karlek:
@Jampi0n, thanks I'll look into those.

@Frotty, I did. I guess that my mpq extractor couldn't apply the patch mpq, and therefore it did not find the audio files of the tavern heroes. 

Sweet, I'll look into it.

PR #41, opened 2018-01-17
Add dialog ext functions and DialogBox wrapper

Jampi0n:
Allows dot-syntax for dialogs.
Package is comparable to other packages of the _handles folder.


IgorSamurovic:
Looks solid. It seems to wrap all functionalities of dialogs (if I am not mistaken). You can possibly extend some functionalities to make it even easier to use or solve some typical dialog issues, but obviously, you don't need to add something that nobody (including yourself) would use. So don't be afraid to add more functionalities if you believe they could make this wrapper even more useful.

Frotty:
Cool 😎 
It would be neat to have some hot doc `/** example */` above at least some of the functions, like `addQuitButton` to elaborate on it's function (does quit refer to quitting the dialog, or quitting the game?) as well as a small package description with a few known pitfalls (dialogs pause the game in SP, but not MP iirc and the user is unable to chat because enter acts as pressing a button on the dialog)

Jampi0n:
Thank you. Yes this is just for better syntax of dialog natives.
I made a more advanced dialog system using two classes DialogBox and DialogButton to make dialogs easier to use.
The idea is to only create the DialogBox  with a title and adding buttons to it with text and a function to run, if that button is clicked.
`	DialogBox db = new DialogBox("Example dialog")
	db.addButton("Button 1", ()-> print("Button 1 clicked") )
	db.display(players[0], true)`
This is probably too advanced for a basic package from the standard library though.

I am no expert of dialogs at all. I simply made this package for the better syntax. I decided to do all dialog natives while I am at it, so I can't say what addQuitButton does right now.
I will investigate on it and add descriptions to the package.

IgorSamurovic:
> This is probably too advanced for a basic package from the standard library though.

Nope, not at all. The standard library isn't about "basic packages", it is about encapsulating functionality in the best way possible. If something's generally useful, it should be part of it. Adding lambda functionality like this is actually very helpful, especially if you write proper hotdoc alongside it. 

So if you ask me, you can add this lambda based functionality, since the users would probably use something like that anyways.

Jampi0n:
Ok, I will try to clean up that code a bit. Getting feedback on it will be very useful for me, as I just started using wurst.
This function works with both lambdas and normal functions, but I agree here lambas are very useful, because most of these functions  are used only in this context anyway.

IgorSamurovic:
I am actually not too sure that you need a wrapper class for this, I think working with the basic dialog should do just fine, especially if you just use a HashMap<button, closure> to work with it.

The rule of the thumb is not to create a wrapper class until you really stand to gain something from it, which in this case, you really don't.

wurst/_handles/Dialog.wurst
```diff
@@ -0,0 +1,23 @@
...
+    DialogClear(this)
+
+public function dialog.destr()
+    DialogDestroy(this)
+
+public function dialog.display(player whichPlayer, boolean flag)
+    DialogDisplay(whichPlayer, this, flag)
+
+public function dialog.setMessage(string messageText)
+    DialogSetMessage(this, messageText)
```
Cokemonkey11: missing eof lf

Cokemonkey11:
I agree that the code is good as-is. Extension functions for natives is almost always useful.

I also think a wrapper API would be nice, though not sure about the design, and definitely not needed to have *instead* of this. A wrapper API should use the extension-functions API.

Jampi0n:
I added the DialogBox class.
I am not happy with the number of data structures I needed to make this rather simple system work.
Especially regarding the code parameter I think it can be improved, so I don't need to create one trigger per button.

wurst/_handles/Dialog.wurst
```diff
@@ -0,0 +1,79 @@
...
+
+    static function removeClickedDialog()
+        button b = GetClickedButton()
+        buttonTriggers.get(b).execute()
+        destroy(dialogBoxes.get(GetClickedDialog()))
+
+    function addButton(string buttonText, integer hotkey, code cb)
+        button b = this.dialogHandle.addButton(buttonText, hotkey)
+        this.buttons.add(b)
+        trigger t = CreateTrigger()
```
Frotty: again, use let and .. and proper names

wurst/_handles/Dialog.wurst
```diff
@@ -20,4 +23,57 @@ public function dialog.display(player whichPlayer, boolean flag)
...
+        t.addAction(cb)
+        buttonTriggers.put(b, t)
+
+    function addButton(string buttonText, code cb)
+        this.addButton(buttonText, 0 , cb)
+
```
IgorSamurovic: Add the action that calls ```removeClickedDialog``` on both ```addButton``` functions. This way the user does not have to use it explicitly. Also make ```removeClickedDialog``` a ```private static function```.

wurst/_handles/Dialog.wurst
```diff
@@ -0,0 +1,79 @@
...
+    function display(player whichPlayer, boolean flag)
+        this.dialogHandle.display(whichPlayer, flag)
+
+    ondestroy
+        while(not buttons.isEmpty())
```
Frotty: use a for each loop, either for from with iterator and using remove, or simple .forEach() then clear the list before destroying.
Dialogs aren't performance critical so we aim for most readable, concise and simple code.

wurst/_handles/Dialog.wurst
```diff
@@ -0,0 +1,79 @@
...
+            button b = buttons.pop()
+            trigger t = buttonTriggers.get(b)
+            buttonTriggers.remove(b)
+            t.clearActions()
```
Frotty: .. use cascade

wurst/_handles/Dialog.wurst
```diff
@@ -0,0 +1,79 @@
...
+            t.destr()
+        destroy(buttons)
```
Frotty: omit brackets

Frotty:
I have mostly marked convention and readability issues for now and not design. In any case it is missing hot doc documentation on the usage and for the functions.

wurst/_handles/Dialog.wurst
IgorSamurovic: Or you can just do 
```
for btn in buttons
    -- do stuff ---
destroy buttons
```

Jampi0n:
Thank you for all the feedback. I tried to adress all the issues as far as I understood them.
I added a little bit hot doc documentation.
I will improve it and reply to some of your comments once I have more time.

wurst/_handles/DialogBox.wurst
```diff
@@ -0,0 +1,67 @@
...
+public class DialogBox
+
+    private static constant HashMap<button, trigger> buttonTriggers = new HashMap<button, trigger>
+    private static constant HashMap<dialog, DialogBox> dialogBoxes = new HashMap<dialog, DialogBox>
+    private constant LinkedList<button> buttons = new LinkedList<button>
+
+    private constant dialog dialogHandle = createDialog()
+    private constant trigger clickDialog = CreateTrigger()
+        ..registerDialogEvent(this.dialogHandle)
+        ..addAction(function removeClickedDialog)
```
Frotty: I only ment the `Create` part.
Move 
```
..registerDialogEvent(this.dialogHandle)
..addAction(function removeClickedDialog)
```

back into constructor.
Change `..addAction(function removeClickedDialog)` to `..addAction(() -> removeClickedDialog())`

wurst/_handles/Dialog.wurst
```diff
@@ -0,0 +1,43 @@
...
+    return DialogCreate()
+
+public function dialog.addButton(string buttonText) returns button
+    return this.addButton(buttonText, 0)
+
+/**Hotkey: use ASCII numbers of the capital letter.*/
+public function dialog.addButton(string buttonText, integer hotkey) returns button
+    return DialogAddButton(this, buttonText, hotkey)
+
+/**If clicked, quits the game*/
```
Frotty: Should be more like "adds a quit button to this dialog. If it is clicked, it ends the game for that player(?)"

wurst/_handles/Dialog.wurst
```diff
@@ -0,0 +1,43 @@
...
+public function dialog.addQuitButton(boolean doScoreScreen, string buttonText) returns button
+    return this.addQuitButton(doScoreScreen, buttonText, 0)
+
+/**If clicked, quits the game
+Hotkey: use ASCII numbers of the capital letter.*/
+public function dialog.addQuitButton(boolean doScoreScreen, string buttonText, integer hotkey) returns button
+    return DialogAddQuitButton(this, doScoreScreen, buttonText, hotkey)
+
+/**removes all buttons from a dialog*/
```
Frotty: capitalize r

wurst/_handles/DialogBox.wurst
```diff
@@ -0,0 +1,67 @@
+package DialogBox
```
IgorSamurovic: Should probably be in ```wurst/util/``` instead of _handles, since it's kinda reserved for primitive, wc3 types.

wurst/_handles/DialogBox.wurst
```diff
@@ -0,0 +1,67 @@
...
+import HashMap
+import Dialog
+import LinkedList
+
+/**DialogBox lets you to create dialogs and associate the buttons with code.*/
+public class DialogBox
+
+    private static constant HashMap<button, trigger> buttonTriggers = new HashMap<button, trigger>
+    private static constant HashMap<dialog, DialogBox> dialogBoxes = new HashMap<dialog, DialogBox>
```
IgorSamurovic: Again, no need to denote the type.
```private static let dialogBoxes = new HashMap<dialog, DialogBox>```
will work just fine.

File: wurst/_handles/DialogBox.wurst [hunk 12]
```diff
@@ -0,0 +1,67 @@
...
+    private static function removeClickedDialog()
+        let btn = GetClickedButton()
+        buttonTriggers.get(btn).execute()
+        destroy dialogBoxes.get(GetClickedDialog())
+
+    function addButton(string buttonText, code cb)
+        this.addButton(buttonText, 0 , cb)
+
+    /**Hotkey: use ASCII number of the capital letter.*/
+    function addButton(string buttonText, integer hotkey, code cb)
```
Frotty: The general consensus here should be to use a closure instead of code, so the user does not have to worry about carrying over data.
Imagine following freehand example (didnt test):
```
enum UpgType
    A
    B
    C
class UpgradeShop
    DialogBox dialogBox = new DialogBox("Select Upgrade")
    unit buyingUnit = null

    construct()
        dialogBox..addButton("A", () -> onUpgrade(UpgType.A))
        ..addButton("B", () -> onUpgrade(UpgType.B))
        ..addButton("C", () -> onUpgrade(UpgType.C))
        
    function onUpgrade(UpgType utype)
        dialogBox..display(buyingUnit, false)
        buyingUnit.addBonus(utype.getBonus())
        buyingUnit = null
       
    function showShop(unit buyer)
        if buyingUnit == null
            buyingUnit = buyer
            dialogBox.display(buyingUnit.getOwner(), true)
```

With the use of closures, the class instance and therefore `buyingUnit` are automatically available inside the lambda expression, which is not the case with code.
In that case the user would have to attach his data manually somehow. 
Using closures would also get rid of triggers completely.

wurst/_handles/DialogBox.wurst
```diff
@@ -0,0 +1,67 @@
...
+    ondestroy
+        buttons.forEach( (button btn) -> begin
+            buttonTriggers.get(btn)
+            ..clearActions()
+            ..destr()
+        end )
+
+        destroy buttons
+        dialogBoxes.remove(this.dialogHandle)
+        this.dialogHandle.clear()
```
Frotty: ..

wurst/_handles/DialogBox.wurst
```diff
@@ -0,0 +1,67 @@
...
+    
+
+    construct(string title)
+        this.title = title
+        this.dialogHandle.setMessage(title)
+        dialogBoxes.put(this.dialogHandle, this)
+
+    private static function removeClickedDialog()
+        let btn = GetClickedButton()
+        buttonTriggers.get(btn).execute()
```
Frotty: `buttonTriggers.get(btn)` could be null. Should check if the button has no attached listener and throw an error if so.

Jampi0n: It can be intended to create buttons without a listener. If you simply want a button to cancel the dialog, no code needs to be executed. 
The same goes for the quit button, though I am not sure it would even be used.

wurst/_handles/DialogBox.wurst
Jampi0n: Oh ok, I was already wondering, if executing functions there makes sense.

What is the benefit of using () -> compared to function in this context?

wurst/util/DialogBox.wurst
```diff
@@ -0,0 +1,84 @@
+package DialogBox
+import HashMap
+import Dialog
+import LinkedList
+
+public interface Callback
```
IgorSamurovic: The name is too generic, especially so since it's a public interface. Since it looks like it is only used internally, you could argue that it can simply not be public, however, considering that it is used as an argument (which can also be created and saved outside this system as variables), it should remain public, but change name, to something like DialogBoxButtonClosure. It looks ugly and overly verbose, but closure names are typically long and descriptive because they both kind of need to be in public space, and need to be sound specific enough in order not to clash with others. Luckily, users rarely need to manually write them down, so they're mostly bound to systems they are part of.

wurst/util/DialogBox.wurst
```diff
@@ -0,0 +1,84 @@
...
+        this.title = title
+
+    function getTitle() returns string
+        return this.title
+
+    ondestroy
+        buttons.forEach( (button btn) -> begin
+            let cb = buttonClosures.get(btn)
+            buttonClosures.remove(btn)
+            if(cb!=null)
```
IgorSamurovic: ```if cb != null```

Jampi0n:
There is still one conceptual problem I don't know how to deal with at the moment.
The DialogBox has the functionality to automatically destroy itself when clicked. This is a good feature in singleplayer mode, but does not work for multiplayer dialogs.

To solve this I could use the `display` function to keep track of the players who can see/click the dialog. Once everyone of them has clicked the dialog, it will be destroyed.
Of course the map maker can use destroy to manually destroy the object at any time.

In general I need to do multiplayer tests of the system to be able to tell if it performs as intended in multiplayer as well.

wurst/util/DialogBox.wurst
```diff
@@ -0,0 +1,84 @@
...
+    
+    /** Adds a quit button to this dialog. If it is clicked, it ends the game for that player.
+    Hotkey: use ASCII numbers of the capital letter. */
+    function addQuitButton(boolean doScoreScreen, string buttonText, integer hotkey, Callback cb)
+        let btn = this.dialogHandle.addQuitButton(doScoreScreen, buttonText, hotkey)
+        this.buttons.add(btn)
+        buttonClosures.put(btn, cb)
+
+    /** Toggles visibility of the DialogBox for a player. Dialogs are invisible by default
+    Dialogs cannot be shown at map initialization */
```
IgorSamurovic: Add dot at the end of sentences, if that's how the whole package is documented. Sadly, we haven't yet made proper standards for how this is done (whether there's a dot or not), but having it consistent on a package level will make it easier to refactor later when standards are decided.

wurst/util/DialogBox.wurst
```diff
@@ -0,0 +1,84 @@
...
+        this.dialogHandle.setMessage(title)
+        dialogBoxes.put(this.dialogHandle, this)
+        this.clickDialog
+            ..registerDialogEvent(this.dialogHandle)
+            ..addAction(function removeClickedDialog)
+
+    private static function removeClickedDialog()
+        let btn = GetClickedButton()
+        let cb = buttonClosures.get(btn)
+        if(cb!=null)
```
IgorSamurovic: No need for braces, ```if cb != null``` will suffice.

wurst/util/DialogBox.wurst
```diff
@@ -0,0 +1,84 @@
...
+        destroy dialogBoxes.get(GetClickedDialog())
+
+    function addButton(string buttonText, Callback cb)
+        this.addButton(buttonText, 0 , cb)
+
+    /** Hotkey: use ASCII number of the capital letter. */
+    function addButton(string buttonText, integer hotkey, Callback cb)
+        let btn = this.dialogHandle.addButton(buttonText, hotkey)
+        this.buttons.add(btn)
+        buttonClosures.put(btn, cb)
```
IgorSamurovic: You can't add a null to a ```LinkedList<T>```, so this will throw an error. Check if ```cb``` is null. Unless, of course, this is intended behavior and adding null is an actual error (but I got a different idea from your argument that a user can willfully pass null). In that case, create additional ```addButton``` functions that take only text, and text and hotkey. Don't be afraid to overload functions many times if it makes the user not have to pass null to accomplish something. Even if the code looks a bit messier in the system, it will look neater in the user code, which is what you should aim for if you have 0-4 parameters like here.

IgorSamurovic:
I guess multiplayer is kind of posing a problem.That'll require additional testing to figure out how to solve, though, since we are generally inexperienced with dialogs, and especially so in multiplayer.

wurst/util/DialogBox.wurst
Jampi0n: Thank you for your insight. I was considering overloading the function, but I was not sure if using null as parameter would be easier.
The problem here is, that there are 3 optional paramters, so this will be a lot of functions if every combination is a new function.
The linked list only contains the buttons, which are never null. Only the hashmap maps buttons to closures and the closure could be null. Can you put null values into hashmaps? At least there was not error message when I just tested it using a null closure.

IgorSamurovic: That's fine I guess, I didn't realize at first that it's a map.

Still, there are really not 3 optional parameters - there are only two. You will always need to name the button. This means that the optional parameters are going to be the hotkey and callback. This makes for a 2 * 2 = 4 functions which is okay to manage.

Jampi0n: Yes, I agree. I wasn't thinking about removing the string paramter, but somehow thought there were 3 other paramters. Probably because the quit button contains one additional paramter.

Frotty:
As we discussed, just let the destroying/closing be done by the user.
Often dialogs are reused anyways.
If you make these changes, I will merge @Jampi0n

Jampi0n:
Changes have been made.
I also improved the documentation and added an example use for the class.
As suggested I overloaded the addButton functions for the optional parameters hotkey and closure.

Frotty:
nice 🍉

PR #42, opened 2018-01-19
Add lightning handle package & hero attribute functions for units

Trokkin:
A simple wrappers on lightning handle using vec3/vec2 and color/colorA packages along with some utility wrappers for hero attributes.


wurst/_handles/Unit.wurst
```diff
@@ -42,6 +42,24 @@ public function unit.addState( unitstate state, real value )
 
 public function unit.addXp(int toAdd, boolean showEyeCandy)
 	AddHeroXP(this, toAdd, showEyeCandy)
+public function unit.getStr(bool includeBonuses) returns int
```
Frotty: Leave empty lines between functions

wurst/_handles/Lightning.wurst
```diff
@@ -0,0 +1,30 @@
...
+	return MoveLightningEx(this, checkVisibility, start.x, start.y, start.z,finish.x, finish.y, finish.z)
+
+public function lightning.setColor(colorA c) returns boolean
+	return SetLightningColor(this, c.red / 255, c.green / 255, c.blue / 255, c.alpha / 255)
+	
+public function lightning.setColor(color c) returns boolean
+	return SetLightningColor(this, c.red / 255, c.green / 255, c.blue / 255, 1)
+
+public function lightning.getColor() returns color
+	return color((255 * GetLightningColorR(this) + .5).toInt(), (255 * GetLightningColorG(this) + .5).toInt(), (255 * GetLightningColorB(this) + .5).toInt())
```
Frotty: should make extension funcs for getRed/getBlue etc. also I think

wurst/_handles/Lightning.wurst
```diff
@@ -0,0 +1,30 @@
+package Lightning
+import NoWurst
+import Vectors
+import Colors
+
+public function addLightning(string codeName, bool checkVisibility, vec2 start, vec2 finish) returns lightning
```
Frotty: "finish" should probably be "end" on these functions.
Also what about adding all the default lightning strings in this package?

Trokkin: I had trouble that "end" is a keyword so I can't use that name.
Well, ok, I'll do it if you mind not something like "_end"

wurst/_handles/Lightning.wurst
Trokkin: I thought them would be bad to use,... ok then.

Trokkin:
This should work.
Also, hasn't nobody missed unit.getMaxMana() apart from me? I wonder how it hadn't been added before.

wurst/_handles/Lightning.wurst
```diff
@@ -0,0 +1,102 @@
...
+		case ManaBurn
+			s = "MBUR"
+		case ManaFlare
+			s = "MFPB"
+		case SpiritLink
+			s = "SPLK"
+	return s
+
+/** Acts as with vec3 version but with z equal to terrain height. */
+public function addLightning(LightningType lType, bool checkVisibility, vec2 start, vec2 _end) returns lightning
```
Frotty: I forgot "end" is keyword. I guess "stop" or "finish" or even "_end" is fine.

Frotty:
does the sorting test fail locally for you too?
Not sure what you mean by object description. If you can't write more in editor I don't think it will work.

Trokkin:
By description I mean extended tooltip. Well, I believe I can, looking onto Dota heroes description. They're definitely larger than what I place in my descriptions.
It can be also that I use Cyrillic alphabet for that, but I feel like I saw maps in Russian that contained pretty much same length descriptions.
Also, do you remember a problem with one legacy map when it crashed because of one description that contained unknown symbol at the end? This seems to be similar to what I have, and I know that default world editor won't make the map crash on loading by any means, so I assume the root of the problem is somewhere in wurst obj editing core.

wurst/_handles/Lightning.wurst
```diff
@@ -0,0 +1,68 @@
+package Lightning
+import NoWurst
+import Vectors
+import Colors
+
+public constant string LIGHTNING_CHAIN_LIGHTNING_PRIMARY = "CLPB"
```
Frotty: omit "string"

Frotty:
@Trokkin yes, you should make a ticket then.
Please fix the last comment and I'll merge this. Thanks.

Frotty:
Sweet 🍭

PR #43, opened 2018-01-29
Added PlayerUtils by TriggerHappy

Kithio:
Decided to write over TriggerHappy's PlayerUtils into Wurst as I didn't see any way to iterate through only Active Players (User, Is Playing). I'm sure other people have use of this as well.

User is meant to be a private class that will only be created once, during map init/load, which goes through and adds all players that are Users and Playing. No player is supposed to be able to get access to creating their own User as it would be fairly pointless for them to do so.

_initialize() exists inside of the class in order to construct the User array and build the LinkedList.

Note: _initialize() would not be needed if "construct()" could be made Protected, which.. for some reason isn't possible

### DIFF
--- a/wurst/util/PlayerUtils.wurst
+++ b/wurst/util/PlayerUtils.wurst
@@ -8,7 +8,6 @@ import LinkedList
 
 User array USERS
 
-/** Get User from  */
 public function player.toUser() returns User
     return USERS[this.getId()]
 
@@ -20,6 +19,7 @@ public class User
     protected player _player
     protected int id
     protected string originalName
+    protected int index
     protected bool isPlaying
 
     private construct(player pl, int playerId)
@@ -61,6 +61,9 @@ public class User
     
     function toPlayer() returns player
         return this._player
+    
+    function toIndex() returns integer
+        return this.index
 
     function isPlaying() returns bool
         return isPlaying
@@ -80,13 +83,17 @@ public class User
 
     protected static function _initialize()
         let t = CreateTrigger()
+        int indexNum = 0
         User u
+
         for i = 0 to 11
             u = new User(players[i], i)
 
             if GetPlayerController(u._player) == MAP_CONTROL_USER and GetPlayerSlotState(u._player) == PLAYER_SLOT_STATE_PLAYING
                 TriggerRegisterPlayerEvent(t, u._player, EVENT_PLAYER_LEAVE)
                 u.isPlaying = true
+                u.index = indexNum
+                indexNum++
                 activeUsers.add(u)
             
             u.originalName = GetPlayerName(u._player)


IgorSamurovic:
@Frotty I told him to PR this so we can discuss how good the solution is and whether it'd be smarter to let the user deal with players instead of the User object (which can still be used internally to store necessary data).

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
...
+*   Credits to TriggerHappy for making PlayerUtils [vJass]
+*    - Link : https://www.hiveworkshop.com/threads/playerutils.278559/
+**/
+
+User array USERS
+
+public function player.toUser() returns User
+    return USERS[this.getId()]
+
+public class User
```
Frotty: I don't really like the class named User if the package is named PlayerUtils

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
+package PlayerUtils
+import LinkedList
+
+/**
+*   Credits to TriggerHappy for making PlayerUtils [vJass]
+*    - Link : https://www.hiveworkshop.com/threads/playerutils.278559/
+**/
+
+User array USERS
```
Frotty: why isn't this a static member of the User class? seems inconsistent

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
+package PlayerUtils
+import LinkedList
+
+/**
+*   Credits to TriggerHappy for making PlayerUtils [vJass]
```
Frotty: Credits can be 1 line, an explanation about this package would be more useful.

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
...
+public class User
+    static constant let activeUsers = new LinkedList<User>
+    static constant player local = GetLocalPlayer()
+    static constant int localId = GetPlayerId(User.local)
+
+    protected player _player
+    protected int id
+    protected string originalName
+    protected int index
+    protected bool isPlaying
```
Frotty: Just initialize with true here instead of in the loop

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
...
+            u = new User(players[i], i)
+
+            if GetPlayerController(u._player) == MAP_CONTROL_USER and GetPlayerSlotState(u._player) == PLAYER_SLOT_STATE_PLAYING
+                TriggerRegisterPlayerEvent(t, u._player, EVENT_PLAYER_LEAVE)
+                u.isPlaying = true
+                u.index = indexNum
+                indexNum++
+                activeUsers.add(u)
+            
+            u.originalName = GetPlayerName(u._player)
```
Frotty: move into constructor

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
...
+
+    protected static function _initialize()
+        let t = CreateTrigger()
+        int indexNum = 0
+        User u
+
+        for i = 0 to 11
+            u = new User(players[i], i)
+
+            if GetPlayerController(u._player) == MAP_CONTROL_USER and GetPlayerSlotState(u._player) == PLAYER_SLOT_STATE_PLAYING
```
Frotty: If a player is bot controlled, isPlaying and index won't be initialized.
So bot players should be either completely excluded or handled here.

Also, use the extension functions `u._player.getSlotState()` etc.

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
...
+                TriggerRegisterPlayerEvent(t, u._player, EVENT_PLAYER_LEAVE)
```
Frotty: use extensionfunc

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
...
+                u.isPlaying = true
+                u.index = indexNum
+                indexNum++
```
Frotty: you don't need indexNum, just use the current list size - unless you also add bots

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
...
+                activeUsers.add(u)
+            
+            u.originalName = GetPlayerName(u._player)
+            USERS[i] = u
+        TriggerAddCondition(t, Filter(function onLeave))
+
+init
```
Frotty: just put the contents of your `_initialize` function in here.

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
...
+        return isPlaying
+
+    static function getNumPlaying() returns int
+        return activeUsers.size()
+
+    static function iterator() returns LLIterator<User>
+        return activeUsers.iterator()
+
+    private static function onLeave() returns bool
+        thistype u = User.fromIndex(GetPlayerId(GetTriggerPlayer()))
```
Frotty: use `let`, use proper variable names that are more than 1 letter, use extension funcs

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
...
+        _player = pl
+        id = playerId
+
+    function getId() returns integer
+        return id
+
+    function getName() returns string
+        return GetPlayerName(this._player)
+
+    function getColoredName() returns string
```
Frotty: already exists as .getColoredName ext func

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
...
+**/
+
+User array USERS
+
+public function player.toUser() returns User
+    return USERS[this.getId()]
+
+public class User
+    static constant let activeUsers = new LinkedList<User>
+    static constant player local = GetLocalPlayer()
```
Frotty: already exists as localPlayer

Frotty:
Yeah, thanks for the PR, and hi @Kithio it's a neat way to learn the wurst conventions. (make sure to star wurst repo :P)
The package is quite rough though and needs some work. You should be using extension funcs, handle bot players, don't define redundant functions/constructs.
Also I don't like the idea of having a leave trigger exclusive to this package, since users also need to handle player leaving in their maps. So perhaps add a closure listener or expose the trigger.

I'm not too much a fan of the "User" lingo tbh. I usually went for something like PlayerData or PlayerWrapper.
Also why are members protected? This would indicate you should/could extend this class for your own needs, but that doesn't seem possible.

Kithio:
Protected variables with get functions is the only way to mimic the functionality of PlayerUtils in Jass which makes use of "readonly" in variables making them oneway properties that can easily be fetched but can't be set to something new from the outside.

If readonly was possible on variables it's definitely  something I would be making use of instead. I'll make adjustments as recommended to the file and push as soon as I'm able

Frotty:
Then they should be simply *private*

wurst/util/PlayerUtils.wurst
Kithio: If you mean move the contents of _initialize into init, then that isn't really possible if I want to keep User as a locked class that can't be created outside its own package. because protected constructors aren't a thing. Making it private and calling it from within itself is only way to make this function the way it was originally designed to.

wurst/util/PlayerUtils.wurst
Kithio: Completely forgot that was a thing until I had use of it earlier (20 mins ago)

wurst/util/PlayerUtils.wurst
Kithio: Creating a User for every player even non-playing ones is to get around "nullpointer" happening if you try to fetch data from a player that isn't actually there. If you think I should adjust then that's alright... And isPlaying is meant to be set to false if bot and index is only set to players that are active because it's used by "toIndex()" which fetches it's position in the ActiveUsers array / list

wurst/util/PlayerUtils.wurst
Kithio: Didn't originally see that, will be removed

wurst/util/PlayerUtils.wurst
Kithio: Only players that are ACTIVE should be  given isPlaying = true, bots / non-active player slot are supposed to have it as false, maybe it needs to be explicitly set there as an original value. I forget if false is set by default on booleans in Jass or not.

wurst/util/PlayerUtils.wurst
Kithio: Comes side-by-side with localId, if you want I'll just erase them both.

wurst/util/PlayerUtils.wurst
Kithio: how about changing the name to PlayerData ? unless you have a better suggestion for a name?

wurst/util/PlayerUtils.wurst
Kithio: Will be adjusted and added

Kithio:
@Frotty They'll be changed to private, They were made protected because I originally had the initializer in "init" making it so I couldn't set the private values outside without a set function which just looks like bad practice. Which contributed to moving initialize inside the class. Along with the fact it should be made so it's a Private constructor so people can't just go make their own User element as it's supposed to be handled as a constant list that only gets reduced if a player leaves.

wurst/util/PlayerUtils.wurst
Frotty: I suppose localId *could* be added to the Player package

wurst/util/PlayerUtils.wurst
Frotty: Then that should be clarified, and inited nonetheless. Relying on default values isn't nice. It could alsobe replaced by an enum.

wurst/util/PlayerUtils.wurst
```diff
@@ -0,0 +1,104 @@
...
+        thistype u = User.fromIndex(GetPlayerId(GetTriggerPlayer()))
+
+        u.isPlaying = false
+        activeUsers.remove(u)
+        return false
+
+    protected static function _initialize()
+        let t = CreateTrigger()
+        int indexNum = 0
+        User u
```
Frotty: merge with u =, use var

Kithio: @Frotty I'm not quite sure what you mean, Do you mean move "User u" into the for loop and f use it with the initializer?

wurst/util/PlayerUtils.wurst
Kithio: Actually looking at the code again, Local and LocalId is used internally to fetch a "User" / PlayerData object from the array so the user can do ClassName.fromLocal() to get their own Data Object from their Local ID. Could change it to use the variables from Player but I think it'd be nicer in code to keep available in the class.

But it's up to you guys how you want it in the end.

SamuelMoriarty:
I'm sorry, but I don't think I quite understand what the purpose of this package is, or even if it should be in the stdlib.

We already have extension functions on player objects to do 90% of what you have here.
The only functionality this package provides that doesn't already exist in another form in stdlib is an iterator of active players, and a cache of the player's original name.

We also have player.isIngame() to test whether a player is an actual human or not, so you should probably use that in place of some checks.

With that said, an iterator to active players could probably belong in Player.wurst, seeing as this package doesn't offer any otherwise new functionality.

As I see it, the purpose of the original PlayerUtils package by TriggerHappy was to provide a class-like syntax for players in vJass and some common utilities, but since Wurst has extension functions, most of this is irrelevant anyway.

In my opinion, this should be merged simply as improvements to the existing player extension funcs.

Some other notes:
You can use (id castTo Player) to force ids of the Player struct to match the ids of the players, that way, you won't have to store the id as a separate value.

You can also use players[id] (or players[this castTo int]) to avoid having to store player objects, since they are already cached in Player.wurst.

Frotty:
Yes, especially in it's current state, it is not fit for the stdlib, which is why it is advised to contact us first, before submitting a request. But since the submitter is new and this is a small submission, it can serve as a learning experience and I didn't just wanna brush it off.

That said, most maps will have a construct similar to this, to access a list of active players, so in theory such a package could be useful, but in this current form it lacks the customization/extendability needed, and will therefore be redundant if users have to setup their own class alongside it (with their own data atatched to players and onLeave listener).

For now I wanted to mainly comment on the code - then, when changes were made, further discuss use of this in stdlib. So he can still use this as a good code snippet example in his own map.

IgorSamurovic:
I mainly told the submitter to upload this so that we can all say our take on it, not just me, as I didn't want to be the only one giving him feedback on code, in order not to overfit his style to my personal preferences.

I also initially thought it'd be more useful as a general wrapper until I actually realized that most of it is already provided by the stdlib, especially since the code like this is kind of a vestige of vjass-type coding, which is something there's no shame in, and I personally had a lot of time getting rid of, so learning why he should not do most of those things is a good lesson in itself.

So this code is not actually useless - the active player iterator is very useful, and a system that manages this iterator is also very useful indeed - but I have come to realize that the User wrapper for player isn't very useful (for reasons already explained), and the iterator could simply contain players directly.

Frotty:
@Kithio will you ever update this?

Kithio:
Functionality of what this would bring, as mentioned, seems to be mostly implemented in other ways through Wurst. Only thing it really doesn't have is a map of active players, but you already provide easy ways to check if a player is actually playing through the stdlib.

Therefor I don't really see the point in updating and reworking this when the missing feature could be implemented into the existing player file in the stdlib.

PR #51, opened 2018-03-04
Add weather effect ids

karlek:


Frotty:
add public import to Assets.wurst ?
Otherwise nice :)

karlek:
Ah, missed that! Fixed in last commit.

Frotty:
👍

PR #52, opened 2018-03-06
assets/AbilityIds

karlek:
The asset mania continues.


Frotty:
Nice 👍 
Soon, we never have to use rawcodes directly ever again!

PR #54, opened 2018-03-08
Add TargetsAllowed to prevent typos

karlek:
Would also be really helpful if we could comment each value so it would be clearer what the differences is between `allies` and `friends` for example.

Any feedback on the choice of suffixing the values with `_t` to avoid the namespace collision?

### DIFF
--- a/wurst/objediting/TargetsAllowed.wurst
+++ b/wurst/objediting/TargetsAllowed.wurst
@@ -2,34 +2,65 @@ package TargetsAllowed
 import NoWurst
 
 public class TargetsAllowed
+	/** Can only target air units. */
 	static constant air          = "air"
+	/** Can only target alive units (non-skeleton). */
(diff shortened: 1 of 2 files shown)


Frotty:
Perhaps public import in `UnitObjEditing`?
I think, ideally, such string lists should never be parameters in object calls, and instead directly provide an abstraction that prevents typos.
Like this, you still have to create the list properly, and there is much room for error.
Of course I will still merge for it's usefulness.

karlek:
I agree with public import in `UnitObjEditing` (fixed in latest commit) and also that passing a list as argument is sub-optimal. This will could be fixed if varargs would be added to wurst, as you've discussed at length in tickets and IRC :)

Frotty:
sweet 🍬 
Can upgrade this in the future then.

PR #55, opened 2018-03-10
Fix addAt function in LinkedList class

vitalexbel:
Current `addAt` function is wrong because it creates the new `LLEntry`, that is connected with the `dummy`. I fixed it. Now the new `LLEntry` is correctly connected with the old `LLEntry` in the place of which it is inserted.
I mean `entry.next` and `entry.prev` relations

### DIFF
--- a/wurst/data/LinkedList.wurst
+++ b/wurst/data/LinkedList.wurst
@@ -215,8 +215,8 @@ public class LinkedList<T>
 
 	/** Adds the given element directly behind the element at the given index */
 	function addAt(T elem, int index)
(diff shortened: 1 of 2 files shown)


Frotty:
Please add a description to your ticket.
Peek does **not** remove the element.
If **addAt** is incorrect, add a test.

vitalexbel:
I did not want to add that commit with Russian commentaries, I'll soon redo PR

Frotty:
Alright 👍 
If you add a Test to `LinkedListTests.wurst` for addAt I will merge.

Frotty:
great 🍭

PR #57, opened 2018-03-20
DamageModify and Illusion packages

Trokkin:


Trokkin:
And I forgot that I changed IDC. Oops.
Anyway, I think it will be good to merge.

PR #58, opened 2018-04-06
Fixed "Attacks prevented" taking string

Trokkin:


Frotty:
You said they are bitsets. Can they be represented with the bitset tuple? or something custom for some safety?

Trokkin:
Reason to leave it int was because there's no bitset option for level closures.
There's an example of illidanChannel ability, it has set/presetOptions with int, but ChannelAbilityPreset package implements bitset wrapping for it. So I'm just supporting what this package is about.

Trokkin:
@Frotty, it's different from armor, movetype etc. in the way that you can't set AttacksPrevented to all the options you need if there's >1 without additional variable. I could do it, but it sounds not right.
So, I guess you just merge this.

Frotty:
i know, hence we have the bitset tuple, right? But ok, I will merge for now.

PR #60, opened 2018-04-06
new DamageModify package

Trokkin:


wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,205 @@
...
+
+/** Deal an exact amount of damage to `target`, crediting kill experience and
+	gold bounty to `attacker` and not triggering any of DamageDetection listeners. Isn't supposed to use outside of the system. */
+function dealCodeDamage(unit attacker, unit target, real damage, bool killingBlow)
+	if damage != 0
+		let hp = target.getHP()
+		disableDamageDetect()
+		if not killingBlow
+			target.setHP(max(hp - damage, .405))
+			UnitDamageTarget(attacker, target, 0. , true, false, ATTACK_TYPE_UNIVERSAL, DAMAGE_TYPE_UNIVERSAL, null) // Is it ever doing something?
```
Frotty: use .damageTarget extension funcs

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,205 @@
...
+						..start(0., function after)
+			else
+				d.target.setHP(life - d.initial - d.amount)
+				d.anounce()
+		else
+			// After all: life += -d.amount + d.initial // - d.initial from event 
+			if life + d.initial - d.amount > d.target.getMaxHP()
+				d.target.addAbility(CHEAT_DEATH_ID)
+				d.target.setHP(life + d.initial - d.amount)
+				getTimer()..setData(d castTo int)
```
Frotty: better use ClosureTimers and get rid of the ugly split functions

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,205 @@
...
+	
+	d.modify()
+
+	let life = d.target.getHP()
+	d.killingBlow = life - .405 < d.amount
+	if d.killingBlow and d.saveTarget
+		d.amount = life - .406
+		d.killingBlow = false
+	
+	if d.killingBlow
```
Frotty: this big branching should be cleaned up

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,205 @@
...
+	function isKillingBlow() returns bool
+		return (not fired and target.getHP() - .405 < amount) or killingBlow
+		
+	private construct()
+	protected static function processEvent() returns Damage
+		let damage = GetEventDamage()
+		if damage == 0
+			return null
+		let d = new Damage()
+		d.id = DAMAGE_ID_NONTRIGGERED
```
Frotty: Why are these not constructor parameters?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,205 @@
...
+import OnUnitEnterLeave
+import ObjectIdGenerator
+import LinkedList
+
+@configurable public constant DAMAGE_ID_NONTRIGGERED = -1
+
+public interface ForDamageCallback
+	function run(Damage d)
+
+LinkedList<ForDamageCallback> damageModifiers = new LinkedList<ForDamageCallback>
```
Frotty: let damageModifiers =

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,205 @@
...
+	/** stores how much health were added */
+	protected real afterDamage = 0.
+	protected bool killingBlow = false
+	private bool fired = false
+	/** check this if damage instance is needed elsewhere before .deal(), though you'll need to delete this manually */
+	int wantSave = 0
+
+	/** applied all modifiers to the damage instance once */
+	function modify()
+		for i in damageModifiers
```
Frotty: if this is never called nested, use static iterator

Trokkin:
I saw @Jampi0n 's PR and instantly merged it into my system, not to mention one new good enumeration.

wurst/closures/ClosureForGroups.wurst
```diff
@@ -56,6 +56,13 @@ public function forUnitsOfPlayer(player p, ForGroupCallback c)
 	GroupEnumUnitsOfPlayer(DUMMY_GROUP, p, filter)
 	popCallback()
 
+/** Executes the given closure for every existing unit */	
+public function forUnits(ForGroupCallback c)
```
Frotty: Requires a better name. Perhaps `forAllUnits`

wurst/closures/ClosureForGroups.wurst
```diff
@@ -56,6 +56,13 @@ public function forUnitsOfPlayer(player p, ForGroupCallback c)
...
+	pushCallback(c)
```
Frotty: you need to push/pop callbacks before each Enum call, not just before/after the loop.
Otherwise if there were another ForGroup closure inside the `forUnits`s one, it wouldn't work properly.

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
+package Illusion
```
Frotty: Why is this still in this PR?

wurst/closures/ClosureForGroups.wurst
```diff
@@ -56,6 +56,13 @@ public function forUnitsOfPlayer(player p, ForGroupCallback c)
 	GroupEnumUnitsOfPlayer(DUMMY_GROUP, p, filter)
 	popCallback()
 
+/** Executes the given closure for every existing unit */	
+public function forUnitsAll(ForGroupCallback c)
+	for i = 0 to bj_MAX_PLAYER_SLOTS - 1
+		pushCallback(c)
+		GroupEnumUnitsOfPlayer(DUMMY_GROUP, players[i], filter)
```
Frotty: Why no `group.` extension function for `GroupEnumUnitsOfPlayer` 😆  ?

wurst/closures/ClosureForGroups.wurst
```diff
@@ -56,6 +56,13 @@ public function forUnitsOfPlayer(player p, ForGroupCallback c)
 	GroupEnumUnitsOfPlayer(DUMMY_GROUP, p, filter)
 	popCallback()
 
+/** Executes the given closure for every existing unit */
```
Frotty: maybe add that this also enums locust units

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,192 @@
+package DamageModify
```
Frotty: So, this is nice, but I'm not entirely happy with it.
For once it misses some documentation. From a `DamageModify` package I expect easy to use `modifyDamage` functions which I can use to modify any damageEvent.
Right now this looks to me as if I have to add a `ForDamageCallback` and check the Damage object passed to it.
Then I also have to manually destroy it.
Can you add some documentation with usage examples? Am I getting this wrong?

I would prefer at least some wrappers to easily manipulate any ongoing event. 

Also can you add some unit tests for this? (if there are missing natives, inform me)

Jampi0n:
You can now use the new native BlzSetEventDamage (real damage) to modify damage more easily.

wurst/util/Illusion.wurst
Trokkin: I told you, this package depends on damageModify. And since I'm working on my map too, I can't wait for it to be added after this pr is merged.

Trokkin: I told you, this package depends on damageModify. And since I'm working on my map too, I can't wait for it to be added after this pr is merged.

Trokkin: I told you, this package depends on damageModify. And since I'm working on my map too, I can't wait for it to be added after this pr is merged.

Trokkin: I told you, this package depends on damageModify. And since I'm working on my map too, I can't wait for it to be added after this pr is merged.

Trokkin: I told you, this package depends on damageModify. And since I'm working on my map too, I can't wait for it to be added after this pr is merged.

Trokkin:
I know, but I keep this compatible with old versions.

wurst/util/Illusion.wurst
Trokkin: But if it is required, then I'll cherrypick/revert this out of this PR.

wurst/_handles/Group.wurst
```diff
@@ -22,6 +23,13 @@ public function group.enumUnitsInRect(rect rec, boolexpr filter)
 public function group.enumUnitsSelected(player p, boolexpr filter)
 	GroupEnumUnitsSelected(this, p, filter)
 
+public function group.enumUnitsOfPlayer(player p, boolexpr filter)
+	GroupEnumUnitsOfPlayer(this, p, filter)
+
+public function group.enumUnitsAll(boolexpr filter)
+	for i = 0 to bj_MAX_PLAYER_SLOTS - 1
```
Cokemonkey11: What value does `bj_MAX_PLAYER_SLOTS` take on before and after 1.29?

What about <1.29 maps being played in 1.29?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+	DamageModify
+	by Trokkin
+
+	API:
+
+	// Pretty much '(Damage d) -> ...'
+	interface ForDamageCallback
+		run(Damage d)
+
+	// these are called before the damage is dealt
```
Cokemonkey11: "These adjust effective damage before any has occurred. Can be called zero or more times."

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+	// can be called multiple amount of times or not called at all
+	addDamageModifier(ForDamageCallback cb)
+	removeDamageModifier(ForDamageCallback cb)
+	
+	// these are called after the damage is dealt
```
Cokemonkey11: "These provide a callback API for after damage has occurred. Only one listener may be present per damage event."

Why the restriction on 1?

What's the use case for supporting both before and after? How much after does it occur - is it using a timer?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+	// are called strictly once per damage instance
+	addDamageListener(ForDamageCallback cb)
+	removeDamageListener(ForDamageCallback cb)
+
+	// a short for 'new Damage(this, target, amount, isSpell, id)..modify()..deal()'
+	unit.damage(unit target, real amount, OPTIONAL bool isSpell = true, int id = 0)
```
Cokemonkey11: I'm a bit confused by `OPTIONAL` here, in particular because you haven't labeled `id` as optional but it has a defaulted value.

Suggest just removing `OPTIONAL` qualifier.

What is `id` for?

What effect does `isSpell` have?

Cokemonkey11: What does `modify` method do? Does it need to be introduced before this docu?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+
+	// a short for '= new Damage(this, target, amount, isSpell, id)..saveInstance()..modify()..deal()'
+	unit.returnDamage(unit target, real amount, bool isSpell, int id) -> Damage
```
Cokemonkey11: I find this pretty weird. Is there a better name than `returnDamage`?

I think I prefer the opposite - `new Damage()..destroyAfter()..modify()..deal()`, and a helper method `unit.oneshotDamage()` or something like that.

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+
+	class Damage
+		
+		// Basic damage event info
+		unit source
+		unit target
+		bool spell
+		real initial
+
+		// .id is used for any unique iterations
```
Cokemonkey11: "unique iterations"?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+		integer id
+		// .amount is used to manipulate damage amount
+		// like, addDamageModifier(d -> d.amount += d.initial * 1.00) adds +100% multiplier onto damage
+		real amount
+
+		// if .saveTarget was called then .target will survive from this damage no matter what
+		saveTarget()
```
Cokemonkey11: Prefer something like `preventDeath`

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+
+		// after damage has happened the Damage object is destroyed if saveInstance() wasn't called
+		saveInstance()
+		// releaseInstance is used to destroy Damage objects. Use this instead of 'destroy'!
```
Cokemonkey11: "Use this instead" - why?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+		releaseInstance()
+
+		// Some obvious wrappers
+		addAmount(real amount)
```
Cokemonkey11: Can you give an example of when this would be valuable, given that your `addDamageModifier` above is already quite succinct?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+		addMultiplier(real mult)
+
+		// .modify() evaluates all damageModifiers on itself
+		// is automatically called on non-triggered damage once
+		// isn't called on manual construction, so user can manually decide how many iterations a damage should have, from 0 to infinity
```
Cokemonkey11: I don't understand this. It sounds like `modify` is the function whose side-effect is to "apply" all attached damage modifiers, but not yet deal damage, is that right?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+		modify()
+
+		// Returns true if the damage is going to kill or already have killed target
```
Cokemonkey11: "or already have killed" I think this means it depends on whether `Damage.deal` has been used yet, right?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+		isKillingBlow() -> bool
+
+		// Damage objects can be created manually to have full control on them from the code
```
Cokemonkey11: I don't think this comment adds any value

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+		construct(unit source, unit target, real amount, bool isSpell, int id)
+		// .deal() runs manually created damage
```
Cokemonkey11: What happens if I never call it?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+		deal()
+
+	// an important note:
+	// for damageListeners .amount is set to actual difference in target's life
+	// so damage won't be more than target's life if it has killed target (or it was saved)
+	// (a unit has 10 hp, it takes 100 damage, for damageModifiers it's 100 damage
+		but for damageListeners it's 10 damage / 10 - .405 damage if .doSaveTarget == true)
```
Cokemonkey11: missing `//` prefix, also prefer ", or" instead of `/` here which can be confused with division.

I think this important note probably should be in hotdoc comments instead, but not sure.

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+
+	Examples:
+	
+	// Leech ability
+		constant int LEECH_ID = compiletime(ABIL_ID_GEN.next())
+		constant int LEECH_DAMAGE = compiletime(GetRandomInt(100, 200)) // just for lulz
+
+		// Implementation #1
```
Cokemonkey11: I'm confused why impl #1 doesn't need a `init` block. In general this could use some clean-up.

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+			d.source.addEffect(Abilities.vampiricAuraTarget, "origin").destr()
+
+			d.releaseInstance()
+
+		// Implementation #2
+		function onLeechEffect2()
+			GetTriggerUnit().damage(GetSpellTargetUnit(), LEECH_DAMAGE, LEECH_ID)
+		
+		init
+			addDamageListener() d ->
```
Cokemonkey11: I think you wanted `d ->` instead the opening `(`

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+					d.source.addHP(d.amount) // Leech effect implementation #2
+					d.source.addEffect(Abilities.vampiricAuraTarget, "origin").destr()
+
+	// Damage resistance buff
+		constant BUFF_DURATION = 5.0
+
+		function onBuffEffect()
+			let target = GetSpellTargetUnit()
+			let effect = target.addEffect(Abilities.voodooAuraTarget, "origin")
+			ForDamageCallback cb = () d ->
```
Cokemonkey11: typo `(d) ->` ?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+*/
+package DamageModify
+import AbilityObjEditing
+import DamageDetection
+import OnUnitEnterLeave
+import ObjectIdGenerator
+import LinkedList
+import ClosureTimers
+
+@configurable public constant DAMAGE_ID_NONTRIGGERED = -1
```
Cokemonkey11: What does `DAMAGE_ID_NONTRIGGERED` do?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+	damageModifiers.remove(cb)
+
+let damageListeners = new LinkedList<ForDamageCallback>
+public function addDamageListener(ForDamageCallback cb)
+	damageListeners.add(cb)
+public function removeDamageListener(ForDamageCallback cb)
+	damageListeners.remove(cb)
+
+/** Deals triggered damage from unit to target.
+	You can specify id if there's any unique id interactions.
```
Cokemonkey11: I still don't understand what `id` is for

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+	Damage instance is saved after being destroyed. */
```
Cokemonkey11: :grimacing: what?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+
+/** Deals triggered damage from unit to target. */
+public function unit.damage(unit target, real amount, bool spell)
+	new Damage(this, target, amount, spell, 0)..modify()..deal()
+/** Deals triggered spell damage from unit to target. */
+public function unit.damage(unit target, real amount)
+	new Damage(this, target, amount, true, 0)..modify()..deal()
+
+// ConvertAttackType(7) is a hidden attack type with some unique properties. For
+// more information see goo.gl/9k8tn .
```
Cokemonkey11: Can you add a caveat "Note that the values in the table at this link are outdated, but the principle still applies."?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+
+	/** check this if you want to save target from damage in all cases */
+	protected bool doSaveTarget = false
+	function saveTarget()
+		doSaveTarget = true
+
+	/** Damage instances are automatically destroyed.
+		Use this to protect them if you need that instances somehow. */
+	function saveInstance()
+		wantSave += 1
```
Cokemonkey11: I'm surprised to see that this is reference counting. Can you give an example of where that's useful?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+
+	/** Releases Damage instance so it can be destroyed. */
+	function releaseInstance()
+		wantSave -= 1
+		if fired and wantSave <= 0
+			destroy this
+
+	protected int wantSave = 0
```
Cokemonkey11: protected values probably all want hotdoc comments?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+	
+	/** stores how much health were added */
```
Cokemonkey11: I don't think this comment is clear. Do you mean "stores the sum of the damage modifiers, after base damage"?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+		let damage = GetEventDamage()
+		if damage == 0
+			return null
+		let d = new Damage(GetEventDamageSource(), GetTriggerUnit(), (damage < 0 ? -damage : damage), damage < 0, DAMAGE_ID_NONTRIGGERED)
+		d.fired = true
+		return d
+
+	/** applies user-created Damage instance. */
+	function deal()
+		if not fired
```
Cokemonkey11: So `deal` just silently doe nothing if it's already been fired? I don't think that's good enough

wurst/_handles/Group.wurst
Trokkin: Do wurst inline these constants?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+			fired = true
+			if target.isAlive()
+				let life = target.getHP()
+				killingBlow = life - .405 < amount
+				if killingBlow and doSaveTarget
+					amount = target.getHP() - .406
+					killingBlow = false			
+				dealCodeDamage(source, target, amount, killingBlow)
+				anounce()
+			if this.wantSave == 0
```
Cokemonkey11: Given that `release` can be called arbitrarily, I suggest checking for `<= 0`

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+				destroy this
+
+/** Deal an exact amount of damage to `target`, crediting kill experience and
+	gold bounty to `attacker` and not triggering any of DamageDetection listeners. KillingBlow parameter decides whether to kill the target. Isn't supposed to use outside of the system. */
+function dealCodeDamage(unit attacker, unit target, real damage, bool killingBlow)
+	if damage != 0
```
Cokemonkey11: What effect does this have for very small epsilon values?

What value does this check have?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+			// would otherwise block all damage. The ATTACK_TYPE_MAGIC case affect
+			// ethereal units, whereas other attack types do not.
+			target.setHP(.405)
+			attacker.damageTarget(target, 1000000., ATTACK_TYPE_UNIVERSAL)
+			attacker.damageTarget(target, 1000000., ATTACK_TYPE_MAGIC)
+		enableDamageDetect()
+
+function handler() returns bool
+	let d = Damage.processEvent()
+	if d == null
```
Cokemonkey11: I don't think this check ever does anything

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+
+	let life = d.target.getHP()
+	d.killingBlow = life - .405 < d.amount
+	if d.killingBlow
+		if d.doSaveTarget
+			d.amount = life - .406
+			d.killingBlow = false
+		else
+			// So the unit should die no matter what. The following line should secure it.
+			d.amount = life // A bit of logic wc3 does not have - damage can't be greater than unit's life.
```
Cokemonkey11: I think this comment is unclear. What you mean is "after a unit has died, any natives that query damage dealt will return the unit's life, and we follow that convention" - but that's still confusing.

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+			dealCodeDamage(d.source, d.target, life, true) // DIE, INSECT!!!
+			d.anounce()
+			return false
+		
+	real rLife
+	bool doTimer
+	if d.spell
```
Cokemonkey11: Are you sure that `spell` is the parameter you wanted to check here..?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,327 @@
...
+	new AbilityDefinitionRunedBracers(DAMAGE_TYPE_CHECK_ID)
+		..setName("Runed Bracer Dummy")
+		..setEditorSuffix("(DamageType)")
+		..setItemAbility(false)
+		..setDamageReduction(1, 2.)
+	new AbilityDefinitionAIhe(CHEAT_DEATH_ID)
+		..setName("Cheat Death Dummy")
+		..setEditorSuffix("(Damage)")
+		..setItemAbility(false)
+		..setHitPointsGained(1, 1000000)
```
Cokemonkey11: Missing EoF LF

wurst/event/DamageType.wurst
```diff
@@ -1,10 +1,11 @@
-// Use this package for finding the damage type in the context of a
-// DamageDetection handler. Typically this means differentiating between spells
-// and basic attacks.
+/** Use this package for finding the damage type in the context of a
+    DamageDetection handler. Typically this means differentiating between spells
+    and basic attacks. 
+    
+    It is deprecated tho, since there's DamageModify package. */
```
Cokemonkey11: Use proper english.

Also I don't think this is a strong enough warning. The key concern is that you definitely shouldn't use both this library and `DamageModify`, right?

wurst/event/DamageType.wurst
```diff
@@ -138,7 +139,7 @@ function handler() returns bool
 
         let sampledLife = tU.getHP() - .405
         if sampledLife >= attemptedDamage and sampledLife <= 2.*attemptedDamage
-            tU.setHP(sampledLife - attemptedDamage)
+            tU.setHP(sampledLife - attemptedDamage + .405)
```
Cokemonkey11: Why did you change this?

wurst/event/DamageType.wurst
```diff
@@ -155,12 +156,8 @@ function handler() returns bool
 
 init
     onEnter() ->
-        GetTriggerUnit()..addAbility(DAMAGE_TYPE_CHECK_ID)..makeAbilityPermanent(DAMAGE_TYPE_CHECK_ID, true)
-
-    forUnitsInRect(bj_mapInitialPlayableArea) iter ->
-        iter..addAbility(          DAMAGE_TYPE_CHECK_ID)
-            ..makeAbilityPermanent(DAMAGE_TYPE_CHECK_ID, true)
-
+        getEnterLeaveUnit()..addAbility(DAMAGE_TYPE_CHECK_ID)..makeAbilityPermanent(DAMAGE_TYPE_CHECK_ID, true)
```
Cokemonkey11: Why did you change this?

wurst/event/DamageType.wurst
```diff
@@ -155,12 +156,8 @@ function handler() returns bool
 
 init
     onEnter() ->
-        GetTriggerUnit()..addAbility(DAMAGE_TYPE_CHECK_ID)..makeAbilityPermanent(DAMAGE_TYPE_CHECK_ID, true)
-
-    forUnitsInRect(bj_mapInitialPlayableArea) iter ->
```
Cokemonkey11: Why did you remove this?

wurst/util/Illusion.wurst
Cokemonkey11: Revert it and use a second branch as a target for your own map.

When changes are made to `trokkin:DamageModify`, just make your map's branch rebase onto `trokkin:DamageModify`

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+import Orders
+import HashMap
+import DamageModify
+import Printing
+import OnUnitEnterLeave
+import AbilityObjEditing
+import InstantDummyCaster
+
+@configurable constant int ILLUSION_ID = 'AILS'
+@configurable constant int REFRESH_COUNT = 30
```
Cokemonkey11: What is REFRESH_COUNT for?

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+import Unit
+import Orders
+import HashMap
+import DamageModify
+import Printing
+import OnUnitEnterLeave
+import AbilityObjEditing
+import InstantDummyCaster
+
+@configurable constant int ILLUSION_ID = 'AILS'
```
Cokemonkey11: What's the value in using a configurable constant instead of compiletime generated value?

wurst/event/DamageModify.wurst
Trokkin: Btw, @Frotty, what kind of unit tests do you need? Because if you're unable to simulate wc3's units the same way the game does, there's nothing to test in damage based package.

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+		..registerAnyUnitEvent(EVENT_PLAYER_UNIT_SUMMON)
+		..addCondition(Condition(() -> begin
+			if IsUnitIllusion(GetSummonedUnit())
+				lastCreatedIllusion = GetSummonedUnit()
+		end))
+		..disable()
+
+	onLeave(() -> begin
+		let ill = Illusion.get(getEnterLeaveUnit())
+		if(ill != null)
```
Cokemonkey11: `if` is not a function

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+			destroy ill
+	end)
+	addDamageModifier() d ->
```
Cokemonkey11: I'm confused. Does this syntax work or not? I think this wanted to be `addDamageModifier(d -> ... )`?

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+		let sourceIllusion = Illusion.get(d.source)
+		if(sourceIllusion != null)
```
Cokemonkey11: `if` is not a function

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+			d.amount *= sourceIllusion.damageGiven
+		let targetIllusion = Illusion.get(d.target)
+		if(targetIllusion != null)
```
Cokemonkey11: `if` is not a function.

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+			d.amount *= targetIllusion.damageTaken
+
+public class Illusion
+	unit u
+	real damageTaken = 1
+	real damageGiven = 1
+
+	static HashMap<unit, Illusion> tb = new HashMap<unit, Illusion>
+	static function get(unit u) returns Illusion
```
Cokemonkey11: I don't understand what this is for.

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+		return tb.get(u)
```
Cokemonkey11: This will do something dirty if the Illusion is destroyed

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+	
+	function setDuration(real time)
+		u.setTimedLife(time)
+
+	construct(unit u)
+		this.u = u
+		tb.put(u, this)		
+
+	static function create(player owner, unit source, vec2 pos, int abilId, int lvl) returns Illusion
```
Cokemonkey11: I appreciate that this library has great error handling, but you need more documentation about error cases. Every function that might return null should document that.

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+		unit u = source.getIllusion(abilId, lvl)
+		if u == null
+			return null
+		if u.isType(UNIT_TYPE_STRUCTURE)
```
Cokemonkey11: I find this reasonable, but should be documented

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+			u.setXY(pos)
+		if owner != null
+			u.setOwner(owner, true)
+		u.setAnimation("stand")
+		return new Illusion(u)
+
+	static function create(player owner, unit source, vec2 pos) returns Illusion
+		return create(owner, source, pos, ILLUSION_ID, 1)
+
+	ondestroy
```
Cokemonkey11: Add a comment "instance is automatically destroyed if the illusion dies."

wurst/util/Illusion.wurst
```diff
@@ -0,0 +1,102 @@
...
+		if u.isAlive()
+			u.kill()
+
+@compiletime function generateObject()
+	new AbilityDefinitionItemIllusion(ILLUSION_ID)
+	..setCooldown(1, 0)
+	..setDurationHero(1, 0)
+	..setDurationNormal(1, 0)
+	..setDamageDealtofnormal(1, 1)
+	..setDamageReceivedMultiplier(1, 1)
```
Cokemonkey11: Missing EoF LF

wurst/_handles/Group.wurst
Cokemonkey11: I don't think wurst compiler would inline BJ constants, no, so maybe it's fine

wurst/util/Illusion.wurst
Cokemonkey11: (needs a comment)

wurst/event/DamageModify.wurst
Trokkin: Hm, do we have any DDS that shout its events more than once? I mentioned the restriction in contrast of damageModifiers, that can be called more than once.

it uses a timer in cases when incoming damage from event takes target's life beyond [.405; MAX_LIFE]. Otherwise, right before the actual damage. 

For example, imagine a spell that deals damage that doubles any effects on it's amount. And an item on it's caster that grants spell lifesteal.
Then any damageModifiers will modify that damage twice, but damageListeners - will be called only once, so you won't get twice as much life from that damage.
Same applies for 'pure' damage that isn't being modified at all, but works with lifesteal.

wurst/event/DamageModify.wurst
Trokkin: 1) object won't be destroyed
2) damage process won't happen at all

wurst/event/DamageModify.wurst
Trokkin: .processEvent() returns null if GetEventDamage == 0

wurst/event/DamageType.wurst
Trokkin: I noticed units lose a bit more health from any damage than they should.
Why would you ever want to remove that .405 points of life from unit?
I didn't test that, but without my fix a 0.01 damage should kill unit with 0.5 hp left

wurst/event/DamageModify.wurst
Trokkin: epsilon values are ignored by '==', not '!=', [real talk](https://www.hiveworkshop.com/threads/real-talk-floats-in-warcraft-3.270579/)

wurst/event/DamageModify.wurst
Trokkin: If there's more than one system that saves instance.

wurst/event/DamageType.wurst
Trokkin: forUnitsInRect is automatically iterated in OnUnitEnterLeave package

Frotty:
Any updates on this trokkin?

Trokkin:
Check last 2 commits, I've removed illusion package and applied cokemonkeys advices.

wurst/event/DamageModify.wurst
```diff
@@ -84,11 +85,15 @@
 
 			d.releaseInstance()
 
+		init
```
Cokemonkey11: this file now has two init blocks?

Cokemonkey11:
Looks like you have made some progress but there are a lot of outstanding questions still.

If your view is that a question is not useful, I think you should at least say that, rather than ignore :)

wurst/event/DamageModify.wurst
Trokkin: No, it's two different examples, and they're both just comments after all

wurst/event/DamageType.wurst
Trokkin: onEnter does not operate with getTriggerUnit
It might work, thought it isn't guaranteed to work, especially if onunitenterleave will ever change somehow.

wurst/event/DamageModify.wurst
Trokkin: See examples.
I don't really know how to comment on that field properly, though mostly I assume that id should correspond to ability that caused the damage.

Trokkin: See examples.
I don't really know how to comment on that field properly, though mostly I assume that id should correspond to ability that caused the damage.

wurst/event/DamageModify.wurst
Trokkin: That's what frotty has requested
Like, addDamageModifier() d -> d.addAmount(100500)
Probably just in case dear user wouldn't want to access .amount field directly

Trokkin: That's what frotty has requested
Like, addDamageModifier() d -> d.addAmount(100500)
Probably just in case dear user wouldn't want to access .amount field directly

wurst/event/DamageModify.wurst
Trokkin: Right. The main purpose is to allow intervention at any stage when dealing damage from code.
For example, you can additionally modify the damage before it is dealt after modify method is called.

wurst/event/DamageModify.wurst
Trokkin: Absolutely.
Before .deal it checks .amount,
After .deal it returns if it has killed the target.

Trokkin:
So, any other thoughts about my PR?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,336 @@
...
+	addDamageListener(ForDamageCallback cb)
+	removeDamageListener(ForDamageCallback cb)
+
+	// a helper for 'new Damage(this, target, amount, isSpell, id)..modify()..deal()'
+	unit.damage(unit target, real amount, OPTIONAL bool isSpell = true, int id = 0)
+
+	// a helper for '= new Damage(this, target, amount, isSpell, id)..saveInstance()..modify()..deal()'
+	unit.returnDamage(unit target, real amount, bool isSpell, int id) -> Damage
+
+	class Damage
```
Frotty: This "vjass style" doc of copying the entire class into a comment for documentation is frankly just terrible.
Document the actual class with hotdoc and just provide valuable information in the package doc on how to get started immediately plus most common usages. Most of the comments here are absolutely superfluous and don't add anything of value.

Frotty:
Systems can be part of stdlib, but the implementation should be lightweight and extendable so the user can build upon it. Here i am limited to damagemodifiers, which lacks I mainly want from a DamageType/Modifier system: To get info about and modify an active damage event.
Also, just one global LinkedLists of callbacks shouts problems to me. What if you have a dozen buffs across many units in the map. You will have 50-100 damage modifiers being executed on every damage event. Pretty sure that will cause some performance/oplimit problems.
So I think this needs at least some way to apply unit local damage listeners/modifiers, otherwise the user has to make a global damagemodifier where he does the unit specification themselves.

Also, something like `Never use both this and DamageType package.` is not advisable to have.

Trokkin:
Aren't damagemodifiers informative and modifiable enough since there's Damage object to work with?
And sure, the user should design his own way to store his modifiers efficiently.

Trokkin:
Updated, implemented unit-unique lists for damage modifiers, replaced that `terrible vjassish` API description with proper hotdocs and made changes to how the system is used. So I ask for a re-review.

> make it so it **can't** happen
I have no idea on how to detect or break non-dependant package. Do you have any?

Trokkin:
So, any comments on this?

wurst/event/DamageModify.wurst
Cokemonkey11: For cases that my question warrants a useful explanation, I expect some changes in code comments

wurst/event/DamageModify.wurst
Cokemonkey11: I think you should remove comments that don't make any sense/aren't meaningful

wurst/event/DamageModify.wurst
Cokemonkey11: I find that a bit weird. Maybe could use a diagram in documentation

wurst/event/DamageModify.wurst
Cokemonkey11: So maybe the docs should say so

wurst/event/DamageModify.wurst
Cokemonkey11: Maybe document that

wurst/event/DamageModify.wurst
Cokemonkey11: Maybe document it

wurst/event/DamageModify.wurst
Cokemonkey11: What value does this check have?

Cokemonkey11:
Can you give a summary of what regression checks you have for this package? I see that you have made lots of changes since the last full review...

wurst/_handles/Group.wurst
Frotty: bj_MAX_PLAYER_SLOTS is not a constant. If it were a constant, it would get inlined, BJ or not.

Trokkin:
@Cokemonkey11

Trokkin:
So I:
- Rewrote API description into hotdocs
- Implemented target/source unique damage listeners/modifiers (line 106)
- Changed the way damage event is modified with 1.29 native (line 309) (and removed the `CHEAT_DEATH` ability)
- Changed the way how user-created Damage instances are dealt (line 240)
- Fixed several misconceptions/bugs I found

Cokemonkey11:
My question is about regression checks - as in, what have you done to ensure those 5 bullet points didn't break something that was previously working?

Regression checks might be unit tests, continuous integration FVs, demo maps, etc. Can you give a summary of that stuff?

Trokkin:
Ah, got it *another language barrier has fallen*

This system was initially developed for my map, where damage modification is used heavily. So I was testing from there, and my changes didn't seem to broke anything.
But, yes, I should make a demo map for this exact system, because of my habit to say "this is working but I never tested".

Are unit tests support units and damage right now? Without it, proper unittesting this system is kind of impossible task.
Also, I didn't understand what continuous integration FV is.

Cokemonkey11:
> Ah, got it (another language barrier has fallen)

It's best I guess to ask when unsure. I noticed that I started using more business talk which is not very helpful :disappointed: 

> This system was initially developed for my map, where damage modification is used heavily. So I was testing from there, and my changes didn't seem to broke anything.
> But, yes, I should make a demo map for this exact system, because of my habit to say "this is working but I never tested".

Actually, it sounds to me like you have tested it, but it will be hard to verify since someone like me will have to use your map. What you want is a MWE that demonstrates that it works correctly and also effectively in unusual circumstances.

> Are unit tests support units and damage right now? Without it, proper unittesting this system is kind of impossible task.

I don't know, Frotty? It's possible to unit-test only the mathematical parts for example, which might lead you to better modularity as well. I'm not asking for unit-tests - just asking for a summary of what regressible testing you do have.

> Also, I didn't understand what continuous integration FV is.

Continuous integration: runs on every commit (jenkins or TravisCI)

FV: Functional verification tests (the next level up from unit tests)

Again, I'm not asking you to write more FV/UT tests - but based on your description so far, I do think you need more demonstrable tests.

Trokkin:
*oh that's a misclick*

So, I will make a test project. But where shall I demonstrate it?

Cokemonkey11:
Here is a demo map: https://github.com/Cokemonkey11/knockback3-demo - that was included in a MR for standard library.

I wonder if Frotty has a better idea now, though

Frotty:
One thing that you might want to prevent, are cycles in modifier responses, like 2 units with a damage deflect upon attack, attacking each other. It would result in 2 modifiers calling each other and the game or thread crashing. But I'm not sure if this belongs into the scope of this package.

Test map will be helpful. I will want to move my maps from DamageMod to this system to verify it working alright, too.

Cokemonkey11:
> It would result in 2 modifiers calling each other and the game or thread crashing.

For reference - In DamageType, the issue is not addressed. Instead, the API calls for the user to use a wrapper function (damageTarget) which sets a static-lifetime variable to "CODE DAMAGE" constant. Does DamageModify have something similar?

Trokkin:
There's Damage.id field which should be used as ability source reference. Though, it's up to user, so they can define reflected_damage_ID and check for it before damage reflection. Just make sure that there won't be any other usage of that id.

Trokkin:
Check this out: https://github.com/Trokkin/DamageModify-demo

Cokemonkey11:
I don't plan to test the demo but the presence of one is a good sign imo

wurst/_handles/Group.wurst
```diff
@@ -14,7 +13,7 @@ public function group.enumUnitsInRange(vec2 pos, real range)
 public function group.enumUnitsInRange(vec2 pos, real range, boolexpr filter)
 	GroupEnumUnitsInRange(this, pos.x, pos.y, range, filter)
 
-public function group.enumUnitsInRect(rect rec)
+public function group.enumUnitsInRect(rect rec)
```
Cokemonkey11: ?

wurst/event/DamageDetection.wurst
```diff
@@ -18,6 +17,8 @@ int funcNext = 0
 trigger current = null
 trigger toDestroy = null
 
+constant filter = Filter(() -> addUnit(GetFilterUnit()))
```
Cokemonkey11: Maybe give this a more descriptive name or some documentation

Frotty:
Okay so remaining points from the demo:
- Rename `ForDamageCallback` to something for sensible
- the `#addXXX` functions should return the closure
- Fix and add hotdoc
- Merge the two init blocks
- The save/release instance stuff is still not ideal

Oh and please update the PR.

wurst/event/DamageType.wurst
```diff
@@ -8,16 +8,11 @@ import DamageDetection
 import OnUnitEnterLeave
```
Halithor: Please undo all the whitespace changes in this file... I've opened a different PR to fix the tabs/spaces thing, but it just bloats this PR.

wurst/event/DamageQueue.wurst
```diff
@@ -0,0 +1,108 @@
...
+	*unit 2 is hit with 50 (25.594 actual) damage, 0.406 left*
+	Start 3
+	End 3
+	*unit 2 is hit with 25 (0 actual) damage, unit 2 lives*
+
+	So the results in this test are different if we take order into consideration.
+
+	This is proved for war3 version 1.30.
+	As you can see, in second case damage processing follows the order given
+*/
```
Halithor: Why not just make the behavior of the queued damage the default? The second case seems like the correct way to implement this entire system. The first one seems confusing to me.

wurst/event/UniqueDamageListening.wurst
```diff
@@ -0,0 +1,135 @@
+package UniqueDamageListening
```
Halithor: This package name is confusing/unclear. Maybe just merge this stuff into the main file? This will probably be a more common use case than the global modifiers.

wurst/event/DamageQueue.wurst
```diff
@@ -0,0 +1,108 @@
...
+
+let queue = new LinkedList<Damage>
+/** This is a replacement for .deal() in case of nested damage call that
+	delays new damage event to run after previous damage was dealt.
+	
+	More info on that in DamageQueue package. */
+public function Damage.dealQueued()
+	if queue.size() < 128
+		if queue.isEmpty()
+			getTimer()..start(0., function queueFlush)
```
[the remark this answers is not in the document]
Trokkin: There's also a part of the event we cannot control - health withdrawal. So there's no way to call Unitdamagetarget inside damage callback and then make it happen after that damage. We can only delay the call.

There's other way to catch an event when the damage is dealt - the more correct one - unit life changed event, but that requires rebuilding a trigger once a damage event which is too much overhead for me.

If you are not consistently using Queued damage, then there's no guarantees, it's right (A thing to mention in hotdoc, thanks). So for example AoE damage calling default UDT would have first all the aoe damage and only then all damage caused by it's callbacks.

wurst/event/UniqueDamageListening.wurst
Trokkin: This was taken out of it just because it took a hell of a lot space in there and basically it's just a simple snippet no one is forced to use.

So you can make your own awesome performance optimization suited right for your needs out of this.

Trokkin: This was taken out of it just because it took a hell of a lot space in there and basically it's just a simple snippet no one is forced to use.

So you can make your own awesome performance optimization suited right for your needs out of this.

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,386 @@
...
+
+	/** If the damage is modified by spell bracers ability. */
+	function isSpell() returns bool
+		return attackType == ATTACK_TYPE_NORMAL
+
+	/** Returns true if the damage is going to kill or already have killed target */
+	function isKillingBlow() returns bool
+		return (state == DamageState.FIRED and killingBlow) or (state != DamageState.FIRED and target.getHP() - .406 < amount)
+
+	function getInitial() returns real
```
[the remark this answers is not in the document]
Trokkin: Those three have different behavior, depending  on your needs you would pick a different one.

wurst/event/DamageQueue.wurst
Trokkin: The first one is the default of wc3. And since this snippet is not something simple that were in the game before *and* since it can be implemented other way, it is separated from the main system.
At least that's position I keep because of how @Frotty sees stdlib - every part of it should be as modular and simple as possible, and this conceptually different thing would make DMod too much complicated in one time.

After all, not everybody plays with thorns, so why would those bother themselves with needless overhead?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,386 @@
...
+
+	/** Damage instances are automatically destroyed.
+		Use this to protect them if you need an instance anyhow.
+		
+		You need then to use `releaseInstance()`
+		in order to destroy the instance.
+
+		This means that the instance will be destroyed
+		when the last `releaseInstance()` is called on it */
+	function saveInstance()
```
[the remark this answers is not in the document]
Trokkin: Why copy if you can just save it?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,386 @@
...
+	function getInitial() returns real
+		return initialAmount
+	/** Returns the actual damage that was withdrawn from the unit.
+		Becomes valid only for listeners and after the event. */
+	function getActual() returns real
+		return actualAmount
+	function getAmount() returns real
+		return amount
+
+	protected static Damage previousDamage = null
```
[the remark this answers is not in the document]
Trokkin: Yep, the only reason is to have Damage.getPrevious instead of getPreviousDamage.

wurst/event/DamageModify.wurst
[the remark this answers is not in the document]
Frotty: Three getters are fine, but being able to use them incorrectly is not.
getActual only working in listeners, and being 0 without and user feedback or errors outside of listeners, is not acceptable.

wurst/event/DamageModify.wurst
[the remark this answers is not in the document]
Frotty: I made the exact same suggestion, maybe take a hint Trokkin?
This save/release stuff is unintuitive, 
> "Use this to protect them if you need an instance anyhow."

Doesn't really explain why I need this.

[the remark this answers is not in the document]
Trokkin: Main reason this exists is because damage has a lot of different data inside of it and making a storage for it and then copy it from event would be a lot of a extraneous code.
And some additional overhead, if anybody cares about that.

Basically these two functions mimic shared ptr behavior. Maybe if I will redo documentation based on this fact it could become more intuitive?

wurst/event/DamageModify.wurst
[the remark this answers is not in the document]
Halithor: I prefer the global function for a static method. @Frotty Do you have an opinion on this?

wurst/event/DamageModify.wurst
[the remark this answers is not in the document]
Halithor: I agree with Frotty. I'm still for simplification of this interface. I think it would make more sense to have `getAmount` return the amount that makes the most sense, given the state of the Damage object.

wurst/event/DamageModify.wurst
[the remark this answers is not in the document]
Halithor: To answer your first question, you copy it to have the guarantee that the original object was cleaned up by the library. Furthermore, removing this will simplify the interface exposed to the user.

As a counter point to your reasoning, the damage object does have a lot of data in it and much of it may not want to be stored by the user of this library. It would be beneficial to them to just grab the data they want and store it themselves.

As another thought on this, the save/release methods of this class don't directly relate to the function of this library, and thus shouldn't be exposed to the user.

wurst/event/DamageQueue.wurst
Halithor: The first one isn't default to WC3, it's default to how you structured your code.

I would argue that rather than every part of the std library being as simple as possible, that we should also prioritize the simplicity to use and the intuitiveness of the usage, even if that means some additional complexity in the implementation.

I think there's a way we can deal with nested damage calls elegantly, without increasing the overhead for the case where there are no nested calls.

wurst/event/DamageType.wurst
Halithor: My PR has been merged into master now.

wurst/event/DamageModify.wurst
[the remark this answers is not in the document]
Trokkin: But there are cases where you could want to know all three amounts at once. Or at least it would be bad to deny accessing information. Maybe `getActual()` should return `getAmount()` before listenening stage... I think having a warning in debug stream and amount returned would be a fitting solution.

wurst/event/DamageQueue.wurst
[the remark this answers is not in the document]
Trokkin: `unit.damageTarget` is not a part of this system, so I can't change its behavior. There are some wrappers that overload it with an additional field, and they can use this method of delaying the damage. Even if I would remove/rewrite native `unit.damageTarget`, there's `UnitDamageTarget` native which is not prohibited to use and I cannot change how it is called.

At least since this system is an optional module, it should work around default way to deal damage and not replace it.

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,386 @@
...
+	private real initialAmount
+	/** What amount of damage was convieved after modification*/
+	private real amount
+	/** Final amount of damage event */
+	protected real actualAmount = 0
+	
+	/** Holds info about damage event state */
+	protected DamageState state = DamageState.CREATED
+	/** It is set to true if the damage, when it was dealt, killed the target */
+	protected bool killingBlow = false
```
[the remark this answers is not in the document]
Cokemonkey11: Maybe `targetWasKilled`? I would expect to see `killedTarget` with type `unit`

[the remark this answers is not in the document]
Trokkin: renamed to killedTarget, still, it is a protected field. There are two functions: `hasKilledTarget` and `isKillingBlow` now.

wurst/event/DamageQueue.wurst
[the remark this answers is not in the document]
Halithor: We could put this check inside of the handler, which would have the same effect as placing it into the `unit.damageTarget()` calls that exist outside this library. 

Any calls to any damage dealing natives should trigger your handler, which could then look at the given global state and place the processing of the event into the queue or start a new queue, depending on the state.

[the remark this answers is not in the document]
Halithor: I could write a little mock of a system like this, if you are interested in a more concrete demonstration of what I'm talking about.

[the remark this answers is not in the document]
Trokkin: Ok, so if we are working only in the handler part, then we have no way to stop/delay event from happening. And I'm talking about what game processes - the target recieves the damage and losses hp and maybe dies with credits to damage source.
We should delay unitdamagetarget call, which is impossible if it's already called.

Or your proposal is to make it a 0 damage event and then re-call it?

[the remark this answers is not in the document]
MediumRareSteak-Hive: Unless you try to damage more than 3+ enemies with a very small interval (0.03 and below), using a trigger should be fine for detecting the after damage event.

When dealing with modified damage, (before the existence of BlzSetEventDamage), try to set the health to a bare minimum such that the original damage deals the killing blow, if it is a killing blow.

Now, here's another potential overhead, asserting how much health would be lost. For small values, simple arithmetic should suffice. But when dealing with large values, you can simulate how much health is lost via SetWidgetLife (just make sure to not to kill the unit this way).

If there is no difference between the former health and the latter (result from SetWidgetLife assertion or damage was initially 0.00), fire the After Damage trigger immediately. Otherwise, leave it be.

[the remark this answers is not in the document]
Trokkin: Any AOE damage or crowded battle are bounded to use this system. Too bad for stdlib.
Difference is the order in which things happen. Currently this system leaves proper damage amount for next damage handlers from ddetection, and it should not be changed, so emulating damage dealing is not an option too.
Actually any workaround seems to be too messy, excluding proper order from the beginning.

Trokkin:
So, I believe I've done all the minor things @Halithor requested, plus some key ones:
- `.modify()` method removed, in it's place there are `.get/setModifyCount()` now. So there is no need to call `.modify()` for any time Damage instance is created, but a possibility to change that behavior.
- `.saveInstance()` is removed also, in its place there is a special parameter in the constructor, which decides whether the damage instance will be destroyed during the event. Wrapper methods are adjusted to not have any return value. So it can only be a user who will consciously decide to save the instance and thus take responsibility for instance clean-up.
- DamageQueue is merged into the main package, but its code definitely needs a review and a crash-testing course.

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+		dealingInstance = null // let no one else retrieve the same instance
+	return d
+
+function handler() returns bool
+	let damage = GetEventDamage()
+	if damage == 0
+		return false
+
+	real damage_ratio = 1
+	var d = retrieve()
```
Frotty: use a proper variable name

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+	return new Damage(GetEventDamageSource(), GetTriggerUnit(), GetEventDamage(), false, false, getDamageType() == DamageType.SPELL ? ATTACK_TYPE_NORMAL : ATTACK_TYPE_CHAOS, null, null, DAMAGE_NONTRIGGERED_DATA, false)
+
+// Wrappers for unit.damageTarget with data attachment
+
+/** Deals triggered damage from unit to target. Additional data can be specified.
+
+	Is a helper for `new Damage(*args*, false)..modify()..deal()`, where last
+	arg specifies whether the instance should not be destroyed during the event.
+	If you need instance after it has been dealt, create it manually. */
+public function unit.damageTarget(unit target, real amount, bool attack, bool ranged, attacktype attackType, damagetype damageType, weapontype weaponType, int data)
```
Frotty: these names should be differentiated so you don't mix them up with the normal damageTarget.

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+	damageModifiers.remove(cb)
+
+public function addDamageListener(DamageListener cb) returns DamageListener
+	damageListeners.add(cb)
+	return cb
+
+public function removeDamageListener(DamageListener cb)
+	damageListeners.remove(cb)
+
+// The wrapper class
```
Frotty: ??

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+		case MODIFYING
+			return "modifying"
+		case LISTENING
+			return "listening"
+		case FIRED
+			return "fired"
+
+	return "error"
+
+Damage dealingInstance
```
Frotty: group with other vars above

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+
+public class Damage
+
+	// Event info
```
Frotty: remove pointless comments

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+				else
+					dequeue()
+		else
+			Log.debug("Damage.deal() called on already dealt damage!")
+
+	protected function invoke()
+		state = DamageState.FIRING
+		let temp = dealingInstance
+		dealingInstance = this
+		Log.trace("deal d" + I2S(this castTo int))
```
Frotty: use toString instead of X2S

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+			Log.debug("Damage.setModifyCount() is used in the wrong damage state:" + state.toString())
+				
+
+// Handler of the event
+
+let queue = new LinkedList<Damage>
+var recursionAmount = 0
+var needTimer = false
+var dequeueRunning = false
+timer nullTimer = null
```
Frotty: you don't attach any data to this timer, can it work with 1 constant one?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+	/** Make actual amount leave target at at least 0.406 hp left */
+	function preventDeath()
+		doPreventDeath = true
+
+	/** Nullifies actual amount no matter what. */
+	function deflect()
+		doDeflect = true
+
+	/** Increases `amount` of the damage by `initialAmount * magnitude` */
+	function addMultiplier(real magnitude)
```
Halithor: I'm not a huge fan of these `addEtc` functions. I'd rather just have the `setAmount` function, and let the users implement multipliers or flat adds as they please. These functions aren't saving enough user effort to warrant their existence.

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+	if d == null or d.getSource() != GetEventDamageSource() or d.getTarget() != GetTriggerUnit()
+		if d != null
+			Log.debug("Damage.handler() was ignored a damage due to incorrect target/source!")
+	else
+		dealingInstance = null // let no one else retrieve the same instance
+	return d
+
+function handler() returns bool
+	let damage = GetEventDamage()
+	if damage == 0
```
Halithor: Is there a reason your explicitly ignore 0 damage events? Could you make this a configurable bool `IgnoreZeroDamageEvents`, rather than make it hardcoded functionality?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+	this.damageTarget(target, amount, false, false, ATTACK_TYPE_CHAOS, DAMAGE_TYPE_UNIVERSAL, WEAPON_TYPE_WHOKNOWS, data)
+
+Damage previousDamage = null
+
+/** Returns the damage instance that was created on last Damage event.
+	For both listeners and modifiers it is previous to callback argument.
+
+	Allows to retrieve the instance right after unit.damageTarget() call,
+	but it is not guaranteed. */
+public function getPreviousDamage() returns Damage
```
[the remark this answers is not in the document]
Trokkin: Until Damage was completely processed by the handler, this function will return an instance that was processed before. In case of wrong order recursion (using native damage functions), it would be previous to root Damage, in case of right order recursion, it would be previous to current Damage according to order.

No guarantees due to the recursion possibility - it would return the last, not the first Damage. And if it is used inside damage handler, then it's return value wouldn't change at all from .damageTarget() call, whether that would be native function or wrapper.

wurst/event/DamageModify.wurst
Trokkin: 1) It was intended, although I can't support that intention with any arguments.
2) Could you suggest a different name?

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+/** If Damage instance wasn't created by one of DamageModify package's functions
+	(or `new Damage(...)`), then Damage.data would be set to this value */
+@configurable public constant DAMAGE_NONTRIGGERED_DATA = -1
+
+/** Maximum amount of damage events that can happen in a single event */
+@configurable public constant RECURSION_TRESHOLD = 32
+
+/** Describes default way to create Damage instance if damage event doesn't
+	correspond to any Damage instance. */
+@configurable function getDamageFromEvent() returns Damage
```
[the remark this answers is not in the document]
Trokkin: This creates new Damage instance, that's why it is different from Blizzard notation. It's not critical since it's a private function, but I would want to hear what @Frotty thinks about it too before changing it.

wurst/event/DamageModify.wurst
```diff
@@ -0,0 +1,462 @@
...
+	function modify(Damage d)
+
+/** These callbacks are used to listen to the damage.
+
+	They are called strictly once per every damage instance, guaranteed,
+	but they are disabled to modify. */
+public interface DamageListener
+	function listen(Damage d)
+
+let damageModifiers = new LinkedList<DamageModifier>
```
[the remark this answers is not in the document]
Trokkin: HashSet does not support multiple insertions of the same modifier. Although I can hardly imagine such a bad case where it won't be done with a simple loop inside the modifier...
... both this and unique damage listening were changed.

wurst/event/DamageModify.wurst
Trokkin: Removed hardcoded functionality, inserted the check into `getDamage_(From)_Event` and also encapsulated Damage data check inside retrieve into a separate configurable function.

wurst/event/DamageModify.wurst
Frotty: I think the funcs can be fine, but they should all just use the `setAmount` function and not be boilerplate.
Also, `Log.debug` is pointless for user feedback. The default level is info, so this just isn't displayed at all 99% of the time.
It should at least be warning level, depending on the func even be `error()` because continuing with the error doesn't make sense.

wurst/event/DamageModify.wurst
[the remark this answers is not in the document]
Frotty: Why is it configurable?
I would probably choose something like `getEventDamageInstance` or so.

wurst/event/DamageModify.wurst
Frotty: I think something like `dealCodeDamage` from DamageType is pretty on point.

wurst/event/DamageModify.wurst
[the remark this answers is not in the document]
Frotty: needs hotdoc adjusments and perhepas same safety?

wurst/event/DamageModify.wurst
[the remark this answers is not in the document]
Trokkin: Sounds better for me.
Configurable because it includes sorting out 0 damage and some default values like `attack type chaos` in case of spell damage. Or someone would like to replace dtype, or to remove non-triggered damage entirely.

wurst/event/DamageModify.wurst
```diff
@@ -5,9 +5,22 @@ import HashSet
 import DamageType
 /*	
 	// Overview
-	The main objective of this system is to have same damage event-related object at all stages of damage event.
-	It implements `Damage` class which contains all arguments of `UnitDamageTarget` native plus one integer for additional attachments.
-	This allows to easily access and modify event data.
+	This system brings the best features from damage systems created in vJass
```
Frotty: Pointless preamble

Frotty:
Thanks again @Trokkin 
I think we learned a lot during the process of this PR regarding API and features.
I will close it now in favor of #212

Cheers 🍻

Trokkin:
You traitor :D
Sure, that one is clearly superior choice for standart library.

PR #61, opened 2018-04-08
Damage detection fix

Jampi0n:
Use player based unit enumeration to also add dead heroes and hidden units to the Damage Detection.

Unhiding units and reviving heroes does not trigger enter region events.


Frotty:
Nice catch 👍

PR #63, opened 2018-04-12
Addition of patch 1.29 handle-related natives.

crojewsk:
Includes wrappers for unit, item and effect native methods.
Changelog for each commit lists particular methods that have been accounted for.

### DIFF
--- a/wurst/_handles/Effect.wurst
+++ b/wurst/_handles/Effect.wurst
@@ -18,58 +18,58 @@ public function addEffect(string path, vec3 pos) returns effect
 	return AddSpecialEffect(path, pos.x, pos.y)
 
 public function effect.setColorByPlayer(player whichPlayer)
-	SetSpecialEffectColorByPlayer(this, whichPlayer)
+	BlzSetSpecialEffectColorByPlayer(this, whichPlayer)
 
 public function effect.setColor(int r, int g, int b)
-	SetSpecialEffectColor(this, r, g, b)
+	BlzSetSpecialEffectColor(this, r, g, b)
 
 public function effect.setAlpha(int alpha)
-	SetSpecialEffectAlpha(this, alpha)
+	BlzSetSpecialEffectAlpha(this, alpha)
 
 public function effect.setScale(real scale)
-	SetSpecialEffectScale(this, scale)
+	BlzSetSpecialEffectScale(this, scale)
 
 public function effect.setPosition(real x, real y, real z)
-	SetSpecialEffectPosition(this, x, y, z)
+	BlzSetSpecialEffectPosition(this, x, y, z)
 
 public function effect.setHeight(real height)
-	SetSpecialEffectHeight(this, height)
+	BlzSetSpecialEffectHeight(this, height)
 
 public function effect.setTimeScale(real timeScale)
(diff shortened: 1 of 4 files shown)


crojewsk:
Looks like the blizzard.j/ common.j are not yet updated on the serves thus native method references cannot be resolved.

Frotty:
@crojewsk Cool, nice stuff.
And yes, I am currently working on 1.29 support for the compiler under this branch https://github.com/wurstscript/WurstScript/tree/1.29

However I'm not sure yet how we should handle backwards compatability.
Probably the natives will just be void when on older patches or something.

crojewsk:
In typical environment one must have backwards compatibility in mind, however, here in war3, once patch hits live each user is basically _obligated_ to update his/her client before enjoying the game.

Of course you can still stick to golden principles and add compatibility to older versions. In such case, "do nothing" or return "default(T)" for void and return type functions respectively should suffice.

Frotty:
I don't think that users are obligated. Quite the opposite actually.
There are big communities in asia and russia that are still using patch 1.24 and 1.26.
And I doubt that all will ever move to latest patches.

Frotty:
Okay, the build still fails because your calls are missing the `Blz` prefix.

Trokkin:
Idk for 1.24, but as far as I know in Russia pretty much everybody uses 1.26a. Though I believe as soon as there will be hacked version, our community will slowly transit on the new patch - mainly due to the awesome 24 players feature.

Frotty:
well it would be a branch, not a fork, but yes. Though I don't like having two branches to maintain, so it might be better to just ignore and warn with lower game versions.
The stdlib shouldn't require the new natives in any existing systems as I see it.

@crojewsk will you fix this?!

wurst/_handles/Effect.wurst
```diff
@@ -16,3 +16,60 @@ public function addEffect(string path, vec2 pos) returns effect
...
+	BlzSetSpecialEffectY(this, y)
+
+public function effect.setZ(real z)
+	BlzSetSpecialEffectZ(this, z)
+
+public function effect.setPositionLoc(location loc)
+	BlzSetSpecialEffectPositionLoc(this, loc)
+
+public function effect.getLocalX() returns real
+	BlzGetLocalSpecialEffectX(this, x)
```
Frotty: 🍨

wurst/_handles/Unit.wurst
```diff
@@ -407,3 +407,93 @@ public function unit.getDefaultAquireRange() returns real
 
 public function unit.getDefaultFlyHeight() returns real
 	return GetUnitDefaultFlyHeight(this)
+
+public function unit.getMaxHP() returns int
```
Frotty: name already exists. Do you know the difference between this and `UNIT_STATE_MAX_LIFE` ?

crojewsk: Returns an int rather than int.

crojewsk:
Adding new player native:
native BlzDecPlayerTechResearched takes player whichPlayer, integer techid, integer levels returns nothing

Frotty:
Nice 🍉 
Now you just need to give your first star to the https://github.com/wurstscript/WurstScript repo 😄

PR #70, opened 2018-05-07
Update to registerPlayerEvent(p, c) call

AlexanderKunze:
Not sure if I read it correctly, but registerPlayerEvent(playerevent p, code c) seems to call registerPlayerEvent(playerevent=p, condition=c, action=null), but I would expect it to call registerPlayerEvent(playerevent=p, condition=null, action=c)

### DIFF
--- a/wurst/event/RegisterEvents.wurst
+++ b/wurst/event/RegisterEvents.wurst
@@ -37,7 +37,7 @@ public function getPlayerUnitEventTrigger(playerunitevent p) returns trigger
 /*****PlayerEvent*****/
 
 public function registerPlayerEvent(playerevent p, code c)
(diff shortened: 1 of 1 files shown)


AlexanderKunze:
I see. I thought it odd that a condition gets added (and thought it might lead to buggy behavior), but if this is interchangeable with an action then I apologize for wasting your time. I hope I'll be able to make some other contribution in the future.

Frotty:
Yea I understand that if the user had control over the trigger, it might matter whether it's an action or condition. But in this case I don't see it having any effect. (Unless you know of something)
> I hope I'll be able to make some other contribution in the future.
Cool, looking forward to it 👍 
Also consider starring the main repository if you enjoy wurst 🍬

PR #72, opened 2018-05-10
Update Unit.wurst to include "remove item by ID"

Ouguiya:
Just added a simple function, unit.removeItemById. Functionality pretty much copied from unit.hasItemById(int itemId). Useful when a user wants to remove an item and doesn't care how many of it there are or where it sits.

### DIFF
--- a/wurst/_handles/Unit.wurst
+++ b/wurst/_handles/Unit.wurst
@@ -280,11 +280,13 @@ public function unit.removeItem(item itm)
 /** Removes the first item of type itemId carried by the unit
 	returns true if an item was found and removed */
 public function unit.removeItemById(int itemId) returns boolean
+	boolean item_found = false
 	for slot = 0 to this.inventorySize()-1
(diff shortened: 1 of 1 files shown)


wurst/_handles/Unit.wurst
```diff
@@ -277,6 +277,12 @@ public function unit.removeAbility(int abil) returns boolean
 public function unit.removeItem(item itm)
 	UnitRemoveItem(this, itm)
 
+public function unit.removeItemById(int itemId)
+	for slot = 0 to this.inventorySize()-1
+		if this.itemInSlot(slot).getTypeId() == itemId
+			UnitRemoveItem(this, this.itemInSlot(slot))
```
Frotty: use extension func here.
Also add hotdoc to explain what exactly it does. (e.g. removing first occurance of the item in unit's inventory).
Maybe even return a boolean if an item has been removed.

Frotty:
Nice, thanks 🍭

PR #76, opened 2018-05-13
Fix vec3.polarProject

RoOoBerT:
Hello,

I think vec3.polarProject was wrong. I will take the simplest example to convince you.
Let's say we have a movable object with speed = 100, angleGround = 0 degrees (facing east) and angleAir = 0 degrees (facing the horizon, parallel to the ground).

vec3.polarProject(100, 0°, 0°) should return (100, 0, 0).
Previous implementation would return (0, 0, 100).


Frotty:
@RoOoBerT Clean your fork. I will not accept PRs with all your old commits.
Also, 0° air points up, therefore you move 100 units z component and the result is (0,0,100).
Why do you think 0° is flat?

PR #77, opened 2018-05-15
Support for Upgrade Object Editing

Halithor:
This adds support for editing Upgrade objects in wurst! Provides methods for all upgrade effects as useful helpers in the UpgradeDefinition. Exposes the basic methods in the W3QDefinition for people who want to live dangerously.


wurst/objediting/SharedEnums.wurst
```diff
@@ -0,0 +1,41 @@
+package SharedEnums
```
karlek: Why not name the file `Race.wurst`?

Halithor: At one point I had more than one Enum in the package (armor type for the set armor defense type effect), but I ended up only sharing race. Perhaps in the future this wouldn't be a bad space for other shared enums to go? Or I can rename it to Race.

Frotty: It would be good to have shared/common objediting stuff in one packge, but the name "SharedEnums" isn't good imo.

wurst/objediting/UpgradeObjEditing.wurst
```diff
@@ -1,47 +1,530 @@
...
+public enum UpgradeClass
+	NONE
+	Armor
+	Artillery
+	Melee
+	Ranged
+	Caster
+
+public enum UpgradeEffectType
+	NONE
```
Frotty: afaik the convention is all uppercase and underscore since these are basically constants.
In any case I wouldn't mix the two.

Frotty:
Amazing ❤️

PR #78, opened 2018-05-16
Add Save/Load functions for hashtables in data/Table

Anchorist:
Adds possibility to save/load hashtables in Table


Frotty:
nice ⚓️

PR #80, opened 2018-05-28
Create Camera.wurst

Ouguiya:
Initial proposal - Includes wrappers for almost all camera functions from common.j and blizzard.j. Functions with (real x, real y) and functions with (location loc) have been replaced by vec2. One function (smartCameraPan) was reimplemented from the blizzard.j function.

### DIFF
--- a/wurst/_handles/Camera.wurst
+++ b/wurst/_handles/Camera.wurst
@@ -8,19 +8,17 @@ public function getCameraBoundMin() returns vec2
 public function getCameraBoundMax() returns vec2
     return vec2(GetCameraBoundMaxX(), GetCameraBoundMaxY())
 
-public function getCameraTargetPositionVec2() returns vec2
+public function getCameraTargetPosition() returns vec2
     return vec2(GetCameraTargetPositionX(), GetCameraTargetPositionY())
 
-public function getCameraTargetPositionVec3() returns vec3
-    let targetPosition = getCameraTargetPositionVec2()
-    return vec3(targetPosition.x, targetPosition.y, GetCameraTargetPositionZ())
+public function getCameraTargetPosition3() returns vec3
+    return getCameraTargetPosition().withZ(GetCameraTargetPositionZ())
 
-public function getCameraEyePositionVec2() returns vec2
+public function getCameraEyePosition() returns vec2
     return vec2(GetCameraEyePositionX(), GetCameraEyePositionY())
 
-public function getCameraEyePositionVec3() returns vec3
-    let eyePosition = getCameraEyePositionVec2()
-    return vec3(eyePosition.x, eyePosition.y, GetCameraEyePositionZ())
+public function getCameraEyePosition3() returns vec3
+    return getCameraEyePosition().withZ(GetCameraEyePositionZ())
 
 public function setCameraPosition(vec2 pos)
     SetCameraPosition(pos.x, pos.y)
@@ -28,8 +26,8 @@ public function setCameraPosition(vec2 pos)
 public function setCameraQuickPosition(vec2 pos)
     SetCameraQuickPosition(pos.x, pos.y)
 
-public function setCameraBounds(vec2 corner1, vec2 corner2, vec2 corner3, vec2 corner4)
-    SetCameraBounds(corner1.x, corner1.y, corner2.x, corner2.y, corner3.x, corner3.y, corner4.x, corner4.y)
+public function setCameraBounds(vec2 bottomLeft, vec2 topRight, vec2 topLeft, vec2 bottomRight)
+    SetCameraBounds(bottomLeft.x, bottomLeft.y, topRight.x, topRight.y, topLeft.x, topLeft.y, bottomRight.x, bottomRight.y)
 
 public function panCameraTo(vec2 pos)
     PanCameraTo(pos.x, pos.y)
@@ -70,7 +68,7 @@ public function camerasetup.getField(camerafield whichField) returns real
 public function camerasetup.setDestPosition(vec2 pos, real duration)
     CameraSetupSetDestPosition(this, pos.x, pos.y, duration)
 
-public function camerasetup.getDestPositionVec2() returns vec2
+public function camerasetup.getDestPosition() returns vec2
     return vec2(CameraSetupGetDestPositionX(this),CameraSetupGetDestPositionY(this))
 
 public function camerasetup.apply(boolean doPan, boolean panTimed)
@@ -109,19 +107,21 @@ public function player.panCameraTo(vec2 pos)
 public function player.panCameraToTimed(vec2 pos, real duration)
     PanCameraToTimedForPlayer(this, pos.x, pos.y, duration)
 
+/** Pans the camera depending on how far pos is from the current
+    camera target:
+    - If the distance is larger than
+      bj_SMARTPAN_TRESHOLD_SNAP, the camera snaps instantenously
+    - If the distance is larger than bj_SMARTPAN_TRESHOLD_PAN, the
+      camera pans over the specified duration
+    - If the distance is smaller than bj_SMARTPAN_TRESHOLD_PAN, the
+      camera doesn't move at all. */
 public function player.smartCameraPan(vec2 pos, real duration)
     if GetLocalPlayer() == this
-        let dist = pos.distanceTo(getCameraTargetPositionVec2())   // DistanceBetweenPoints(locA, locB)
-
+        let dist = pos.distanceTo(getCameraTargetPosition())
         if dist >= bj_SMARTPAN_TRESHOLD_SNAP
-            // If the user is too far away, snap the camera.
             panCameraToTimed(pos, 0.0)
-
         else if dist >= bj_SMARTPAN_TRESHOLD_PAN
-            // If the user is moderately close, pan the camera.
             panCameraToTimed(pos, duration)
-
-            // Else: User is close enough, so don't touch the camera.
 
 public function player.setCinematicCamera(string cameraModelFile)
     SetCinematicCameraForPlayer(this, cameraModelFile)
--- a/wurst/_handles/Camera.wurst
+++ b/wurst/_handles/Camera.wurst
@@ -1,6 +1,8 @@
 package Camera
 import NoWurst
 import public Vectors
+import Player
+import TerrainUtils
 
 public function getCameraBoundMin() returns vec2
     return vec2(GetCameraBoundMinX(), GetCameraBoundMinY())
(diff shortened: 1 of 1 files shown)


wurst/_handles/Camera.wurst
```diff
@@ -0,0 +1,166 @@
...
+    return vec2(GetCameraBoundMinX(), GetCameraBoundMinY())
+
+public function getCameraBoundMax() returns vec2
+    return vec2(GetCameraBoundMaxX(), GetCameraBoundMaxY())
+
+public function getCameraTargetPositionVec2() returns vec2
+    return vec2(GetCameraTargetPositionX(), GetCameraTargetPositionY())
+
+public function getCameraTargetPositionVec3() returns vec3
+    let targetPosition = getCameraTargetPositionVec2()
```
Frotty: -> getCameraTargetPositionVec2().withZ(GetCameraTargetPositionZ())

wurst/_handles/Camera.wurst
```diff
@@ -0,0 +1,166 @@
...
+import NoWurst
+import public Vectors
+
+public function getCameraBoundMin() returns vec2
+    return vec2(GetCameraBoundMinX(), GetCameraBoundMinY())
+
+public function getCameraBoundMax() returns vec2
+    return vec2(GetCameraBoundMaxX(), GetCameraBoundMaxY())
+
+public function getCameraTargetPositionVec2() returns vec2
```
Frotty: Don't end function names with "vec", it's the default. Position is enough.

wurst/_handles/Camera.wurst
```diff
@@ -0,0 +1,166 @@
...
+    let eyePosition = getCameraEyePositionVec2()
+    return vec3(eyePosition.x, eyePosition.y, GetCameraEyePositionZ())
+
+public function setCameraPosition(vec2 pos)
+    SetCameraPosition(pos.x, pos.y)
+
+public function setCameraQuickPosition(vec2 pos)
+    SetCameraQuickPosition(pos.x, pos.y)
+
+public function setCameraBounds(vec2 corner1, vec2 corner2, vec2 corner3, vec2 corner4)
```
Frotty: use proper names for the corner points, i.e bottomleft, topRight, etc.

wurst/_handles/Camera.wurst
```diff
@@ -0,0 +1,166 @@
...
+    SetCameraBounds(corner1.x, corner1.y, corner2.x, corner2.y, corner3.x, corner3.y, corner4.x, corner4.y)
+
+public function panCameraTo(vec2 pos)
+    PanCameraTo(pos.x, pos.y)
+
+public function panCameraToTimed(vec2 pos, real duration)
+    PanCameraToTimed(pos.x, pos.y, duration)
+
+public function panCameraToWithZ(vec2 pos, real zOffsetDest)
```
Frotty: why not overload `panCameraTo` with vec3?

wurst/_handles/Camera.wurst
```diff
@@ -0,0 +1,166 @@
...
+public function player.setCameraPosition(vec2 pos)
+    SetCameraPositionForPlayer(this, pos.x, pos.y)
+
+public function player.panCameraTo(vec2 pos)
+    PanCameraToForPlayer(this, pos.x, pos.y)
+
+public function player.panCameraToTimed(vec2 pos, real duration)
+    PanCameraToTimedForPlayer(this, pos.x, pos.y, duration)
+
+public function player.smartCameraPan(vec2 pos, real duration)
```
Frotty: Move single line comments into hotdoc and explain what the func does

wurst/_handles/Camera.wurst
Ouguiya: Great idea!  Implemented.

wurst/_handles/Camera.wurst
Ouguiya: I actually thought about it, but if I understand the function correctly, this does not pan to a specific point in 3D space. It pans to a 2D space and raises or lowers the camera by the amount or zooms the camera in or out (not sure which, but I assume this is why it's real zOffsetDest). The original function signature is: "native PanCameraToTimedWithZ takes real x, real y, real zOffsetDest, real duration returns nothing"

So here, I am worried that using vec3 might give the wrong impression.

wurst/_handles/Camera.wurst
Ouguiya: Done

wurst/_handles/Camera.wurst
Ouguiya: Changed getCameraTargetPositionVec2 to getCameraTargetPosition and getCameraTargetPositionVec3 to  getCameraTargetPosition3
Same with CameraEye

wurst/_handles/Camera.wurst
Trokkin: It does pan to Z, meaning if there would be duration, then camera movement will be smooth. Yes, Z offset will stay and yes, it is not absolute Z, because wc3's camera is always relative to terrain height. But it's still a certain position, which can be translated to absolute 3D coords by adding terrain height to Z.
But in wurst, vec3 is used in two ways: in absolute 3D space and relatively to terrain height. So the vec3 for your function will be the second type, and so there won't be any wrong impressions.

And even if I'm not right, you can forestall that impression by a hotdoc describing that feature.

Ouguiya: @Trokkin I'm a bit confused by your explanation here. The description for one of these types of functions in GUI reads: "Pan camera to (Location) with height x.xx above the terrain over y.yy seconds. The camera will not drop below the terrain height during its camera path."

What would I need to do to achieve the same thing in Camera.wurst?

i.e., if I wrote:

```
public function panCameraTo(vec3 pos)
    PanCameraToWithZ(pos.x, pos.y, pos.z)
```

I'm not sure this would produce the correct result. What would I need to do here?

Trokkin: actually you could see in what JASS code that GUI functions are translated (they're actually BJ functions from Blizzard.j, btw) and look for anything suspicious.
As I know, the result would be correct, if you mean pos.z as the height *above terrain*, which is still a correct usage for vec3 as I explained.

Ouguiya: Unfortunately, that doesn't exactly help here. Here is the JASS code:

```
function PanCameraToTimedLocWithZForPlayer takes player whichPlayer, location loc, real zOffset, real duration returns nothing
    if (GetLocalPlayer() == whichPlayer) then
        // Use only local code (no net traffic) within this block to avoid desyncs.
        call PanCameraToTimedWithZ(GetLocationX(loc), GetLocationY(loc), zOffset, duration)
    endif
endfunction
```

And PanCameraToTimedWithZ is a native, so I can't see what it does.

Now if you tell me that vec3.z component automatically defaults to "height above terrain", then it should be fine. Otherwise, I'm not sure how to force vec3 to use terrain height instead of absolute.

Frotty:
You should make a ticket for the overloading issue that stops the build from succeeding, as we discussed :)

wurst/_handles/Camera.wurst
Trokkin: So, the description should apply 1:1 to that native, as it describes a simple wrapper without any hidden logic.

I told you, that's uncertain: vec3.z is used in both ways as relative and as absolute height, depending on function. See `package Unit` and how vec3 is used there, if you're curious or to find out how to convert back and forth absolute to relative heights. Exactly you want to see getTerrainZ() usage.

Cokemonkey11: Can I ask you guys to make sure you document this in lep/jassdoc?

wurst/_handles/Camera.wurst
```diff
@@ -0,0 +1,166 @@
...
+public function camerasetup.applyForceDurationWithZ(real zDestOffset, real forceDuration)
+    CameraSetupApplyForceDurationWithZ(this, zDestOffset, forceDuration)
+
+public function camerasetup.apply(boolean doPan, player whichPlayer, real duration)
+    CameraSetupApplyForPlayer(doPan, this, whichPlayer, duration)
+
+public function camerasetup.getFieldSwap(camerafield whichField) returns real
+    return CameraSetupGetFieldSwap(whichField, this)
+
+public function player.cameraSetupApply(boolean doPan, camerasetup whichSetup, real duration)
```
Cokemonkey11: I'm not really in love with some of these `player` extension-functions.

wurst/_handles/Camera.wurst
```diff
@@ -0,0 +1,166 @@
...
+public function player.setCameraPosition(vec2 pos)
+    SetCameraPositionForPlayer(this, pos.x, pos.y)
+
+public function player.panCameraTo(vec2 pos)
+    PanCameraToForPlayer(this, pos.x, pos.y)
+
+public function player.panCameraToTimed(vec2 pos, real duration)
+    PanCameraToTimedForPlayer(this, pos.x, pos.y, duration)
+
+/** Pans the camera depending on how far pos is from the current
```
Cokemonkey11: Is this safe in multiplayer?

wurst/_handles/Camera.wurst
```diff
@@ -0,0 +1,166 @@
...
+      camera doesn't move at all. */
+public function player.smartCameraPan(vec2 pos, real duration)
+    if GetLocalPlayer() == this
+        let dist = pos.distanceTo(getCameraTargetPosition())
+        if dist >= bj_SMARTPAN_TRESHOLD_SNAP
+            panCameraToTimed(pos, 0.0)
+        else if dist >= bj_SMARTPAN_TRESHOLD_PAN
+            panCameraToTimed(pos, duration)
+
+public function player.setCinematicCamera(string cameraModelFile)
```
Cokemonkey11: Is string parameter right here? I guess yes, but maybe add some hotdoc.

wurst/_handles/Camera.wurst
Ouguiya: These are the functions that GUI uses, which basically do stuff for the local player (and nothing else). If they weren't safe in Multiplayer, practically any GUI-coded map that alters the camera would desync.

wurst/_handles/Camera.wurst
Ouguiya: If you want me to change anything, you will have to tell me what exactly bothers you about them.

Cokemonkey11: I don't think I'm in a position to demand changes, but I wonder if someone else agrees that `player` extension-functions that are specific to `camerasetup` or otherwise `camera` handling might belong in a separate package that is imported explicitly.

wurst/_handles/Camera.wurst
Cokemonkey11: I'm somewhat confident that the function using SMARTPAN_THRESHOLD can cause a desync in multiplayer, and indeed, a number of GUI functions can cause desyncs.

Frotty: Is `bj_SMARTPAN_TRESHOLD_PAN` modified by gameplay settings or how do you explain using a constant causing desyncs?

wurst/_handles/Camera.wurst
Frotty: I think it's fine as long as this package is an auto import, which it should be.

Trokkin: eh, I feel like it should be camerasetup's extension func instead of player's

wurst/_handles/Camera.wurst
Cokemonkey11: I don't know. I think it's a good strategy to rely on what lep/jassdoc says until proven otherwise. If those BJ functions aren't documented as breaking stuff then we're actually acting as good members of the community by using them, even if that means we find bugs

wurst/_handles/Camera.wurst
Cokemonkey11: Frotty my point about using a separate package imported explicitly is the opposite of what you just said.

In particular, I think `Player(3).panToTimed(vec2(0., 0.), 0.)` is weird to have on auto-import.

Ouguiya: @Trokkin That already exists. I made both versions (for this function and a few others). The one you are looking for is camerasetup.apply, where you specify for which player to do this.

@Cokemonkey11 The function you just mentioned doesn't seem to exist? player.panToTimed isn't in the file. The function is called player.panCameraToTimed, which to me makes sense: Player(3).panCameraToTimed(pos, duration) = For Player 3, pan camera to pos timed over duration.

Frotty: Just saying "it's weird" doesn't help, coke. I think it's totally fine, and I don't see what your problem is. When thinking naturally about this occuring in code, applying a camera position almost always happens together with other player related context. E.g. selecting a unit, adding some resources, etc. and in the same way I would expect to change the player's camera. 
Camerasetup is seldom used, and calling an extension func on it wouldn't allow for cascading. Though it should still be available to the developer, so I think having both is a good idea.
But perhaps they should rely on a shared function to ensure they remain the same.

Cokemonkey11: I didn't know anyone cared why I think it's weird :D actually I just mentioned it in case someone else felt the same.

The reason I think it's weird is that a `player` represents some abstract game/model related interface. Player's methods usually involve manipulating a player's properties in the context of the game. Camera and `camerasetup` is an abstraction around a *view* of the game playing field. Specifying a player when interacting with them only ever impacts the scope over which those views are manipulated.

In other words, I think of the built-in free-function camera API like:

- camera free functions are tools for changing a view over the game terrain
- camera free functions sometimes (always? does it matter?) take a `player` as a means for limiting that scope
- `player` in the sense of the model as related to the game is otherwise not impacted

Anyway, if I really hated this idea, I would be shouting, but I don't, so I just say "seems weird"

Frotty:
Looks good to me now @Cokemonkey11

Cokemonkey11:
LGTM too

PR #81, opened 2018-06-01
Addition of missing and utility unit & item natives.

crojewsk:
Following extension methods have been defined:

- public function unit.dropItemPoint(item itm, real x, real y) returns boolean
- public function unit.dropItemSlot(item itm, int slot) returns boolean
- public function unit.dropItemTarget(item itm, widget target) returns boolean
- public function unit.useItem(item itm) returns boolean
- public function unit.useItemPoint(item itm, real x, real y) returns boolean
- public function unit.useItemTarget(item itm, widget target) returns boolean

- public function item.isAlive() returns boolean
- public function item.isPickupable() returns boolean

### DIFF
--- a/wurst/_handles/Item.wurst
+++ b/wurst/_handles/Item.wurst
@@ -65,9 +65,11 @@ public function item.isSellable() returns boolean
 public function item.isPawnable() returns boolean
 	return IsItemPawnable(this)
 
+/** Verifies whether item is not destroyed and has a valid type id. */
 public function item.isAlive() returns boolean
 	return this.getTypeId() != 0 and GetWidgetLife(this) > .405
 
+/** Verifies whether item can be picked up by a unit by validating its life and ownership status. */
 public function item.isPickupable() returns boolean
 	return this.isAlive() and not this.isOwned() and this.isVisible()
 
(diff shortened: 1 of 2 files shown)


Frotty:
use vec2 where u are using x y

wurst/_handles/Item.wurst
```diff
@@ -65,6 +65,12 @@ public function item.isSellable() returns boolean
 public function item.isPawnable() returns boolean
 	return IsItemPawnable(this)
 
+public function item.isAlive() returns boolean
```
Frotty: add hotdoc for these custom functions please.

crojewsk:
I hope next time I get those two comments of yours in a single one ;X

wurst/_handles/Item.wurst
```diff
@@ -65,6 +65,14 @@ public function item.isSellable() returns boolean
 public function item.isPawnable() returns boolean
 	return IsItemPawnable(this)
 
+/** Verifies whether item is not destroyed and has a valid type id. */
```
Frotty: I believe "verify" is the wrong word to use in both of these functions. 
They don't verify anything. Better change to something like:
`Returns true if the item is not destroyed and has a valid type id`
`Returns true if the item can be picked up by a unit at this moment`

Frotty: I still think `isPickupable` is a bit ambiguous to an item being pickupable by units in general, instead of being pickuable in its current state : )
But I guess it's fine.

Frotty:
👍

PR #82, opened 2018-06-01
Add missing extension functions + prop window fix

Jampi0n:
Added missing getCurrentValue functions for functions that already had getDefaultValue: PropWindow, TurnSpeed, AcquireRange

Added functions with angle as parameter/return value for prop window functions.
Fixed a problem with getDefaultPropWindow:
The native returns the value in degrees, so it has to be converted to radians first. Usually all natives use radians.

Typo: AcquireRange

### DIFF
--- a/wurst/_handles/Unit.wurst
+++ b/wurst/_handles/Unit.wurst
@@ -170,9 +170,11 @@ public function unit.getPos3Zero() returns vec3
 public function unit.getPos3with(real z) returns vec3
 	return vec3(this.getX(), this.getY(), z)
 
+/** Returns the prop window of the unit in radians.*/
 public function unit.getPropWindow() returns real
 	return GetUnitPropWindow(this)
 
+/** Returns the prop window of the unit as angle.*/
 public function unit.getPropWindowAngle() returns angle
 	return angle(GetUnitPropWindow(this))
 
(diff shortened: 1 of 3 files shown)


Frotty:
if the native returns it, an implicit conversion is even worse imo.
The getters should also return angle tuples.

Jampi0n:
The problem is, that GetUnitDefaultPropWindow uses degrees and GetUnitPropWindow uses radians, which makes it inconsistent.
There are getters that use angle, so they will be safe. I can change the functions that use real to return the same as the native and add hotdoc, so the user knows whether it is degrees or radians.
Similar to how it was done with facing angle, I used getPropWindow and getPropWindowAngle.

Trokkin:
+1 for angle tuples, as far as we have them as standart. I think there shouldn't be a single function across the entire stdlib that takes or returns angle as a real instead of the tuple.

Jampi0n:
Ok I will remove the functions that return real then.
As I already said, I did it like this, because it's the same with getFacing and getFacingAngle.
Also regarding the latest commit. The name for the function is quite long, if you know a better name, let me know.

Trokkin:
Hm, maybe you could do an overload for old function with a flag that decides whether to save ratio or not.

crojewsk:
public function unit.setMaxHp(int hp, boolean keepRatio)
public function unit.setMaxMana(int mana, boolean keepRatio)

That's sound about right.

Frotty:
Yes, I agree. We want angle safety everywhere so the user never has to worry about mixing up radians/degrees again.

Jampi0n:
Only uses angle type for functions that use angles. Two old functions that still used reals have been annotated with deprecated.
As suggested setMaxHp/Mana was overloaded with a boolean keepRatio. Wurst's optimizer will remove the condition, as the boolean will be constant in most cases.

Frotty:
Yea looks good now 🍺

PR #84, opened 2018-06-06
Update Unit.wurst

Ouguiya:
Added functions to set the exploding status of a unit and to kill a unit by exploding it.


Frotty:
Thanks 🍬

PR #87, opened 2018-06-11
Added more Effect functionality

eGlint:
- Optimized setPos(vec2) and setPos(vec3).
- The yaw, pitch, roll of special effects can now use angle tuples.
- The color of special effects can now use color and colorA tuples.
- Implemented addEffect to destructables and widgets.
- Created EffectUtils. It contains the EffectEx class which can retrieve the model name, attachment, yaw, pitch, roll, scale, etc. of the special effect. Something that the ordinary Effect package cannot do. It does not support BlzSetSpecialEffectTime since EffectEx cannot track if the special effect is destroyed.


Cokemonkey11:
Looks pretty good to me. Maybe squash?

eGlint:
I apologize, I'm not used to git. How do I squash my commits?

Cokemonkey11:
If you don't know how, it isn't necessary - Frotty can squash on merge.

(But if you google it, you can learn to do it manually)

Frotty:
good stuff, but build is failing. Is it another overloading issue?

eGlint:
Wow, it now works! Thank you!

About the addEffect function for the Unit and Destructable packages, I removed it and placed it into the Widget package instead. It only broke one package in the standard library which is the Fx package. The error occurred due the usage of unit.addEffect and it can be easily fixed by importing the Widget package. 

From my tests, it only breaks from packages that uses import NoWurst and unit.addEffect/destructable.addEffect. Like what I did in the Fx package, it can be easily fixed by importing the Widget package..

It leaves me a question: should we remove the addEffect function from the Unit and Destructable packages? Keep it? Deprecate it?

wurst/_handles/Destructable.wurst
```diff
@@ -68,7 +68,7 @@ public function destructable.setOccluderHeight(real height)
 
 public function destructable.getName() returns string
 	return GetDestructableName(this)
-
+
```
Frotty: shouldnt be here

wurst/_handles/Effect.wurst
```diff
@@ -76,16 +77,35 @@ public function effect.getLocalZ() returns real
...
+	return vec2(BlzGetLocalSpecialEffectX(this), BlzGetLocalSpecialEffectY(this))
+
+public function effect.setColor(color color)
+	BlzSetSpecialEffectColor(this, color.red, color.green, color.blue)
+
+public function effect.setColor(colorA color)
+	BlzSetSpecialEffectColor(this, color.red, color.green, color.blue)
+	BlzSetSpecialEffectAlpha(this, color.alpha)
+
+public function effect.setOrientation(angle yaw, angle pitch, angle roll)
```
Frotty: IIRC with latest patch the yaw/pitch/roll functions still have issues and might produce unexpected behaviour. Ideally add some hotdoc explaining what it does (rotating around which axis, whether it works or not)

wurst/_handles/Unit.wurst
```diff
@@ -26,9 +26,6 @@ public function createUnitZ(player p, int unitId, vec3 pos, angle facing) return
 public function unit.addAbility(int abil) returns boolean
 	return UnitAddAbility(this, abil)
 
-public function unit.addEffect(string fx, string attachment) returns effect
```
Frotty: Before making breaking changes like this, deprecate the function.

wurst/util/EffectUtils.wurst
```diff
@@ -0,0 +1,135 @@
...
+
+        construct(string modelName, widget target, string attachment)
+                initialze()
+                fx = AddSpecialEffectTarget(modelName, target, attachment)
+                this.modelName = modelName
+                this.target = target
+                this.attachment = attachment
+
+        private function initialze()
+                yaw = angle(0)
```
Frotty: these should be merged with the declarations, making this function superfluous

wurst/util/EffectUtils.wurst
```diff
@@ -0,0 +1,135 @@
+package EffectUtils
+import NoWurst
+import Effect
+import Vectors
+import Colors
+import Player
+
+public class EffectEx
```
Frotty: I'm not a fan of the name 'Ex' name and the package name, and, this should be another PR.
Please keep PRs as short as possible and encapsulated, for faster review and merge.

eGlint: Sure, I'll keep that in mind and I'll remove this package in this PR.

wurst/_handles/Effect.wurst
eGlint: I did a test with the yaw/pitch/roll. Based on my test, BlzSetSpecialEffectPitch is the only function having issues when it reaches around 90 to 270 degrees yet the pitch in BlzSetSpecialEffectOrientation seems to work fine. 

Should we keep BlzSetSpecialEffectPitch for setPitch and just add a warning? Or replace it with BlzSetSpecialEffectOrientation until the problem is fixed by Blizzard?

Cokemonkey11: Better set a warning I think

wurst/_handles/Effect.wurst
```diff
@@ -95,15 +95,31 @@ public function effect.setColor(colorA color)
...
 
+/**
+Change the x-axis of the special effect.
+*/
 public function effect.setYaw(angle yaw)
 	BlzSetSpecialEffectYaw(this, yaw.radians())
 
+/**
+Change the y-axis of the special effect.
+Warning: As of 1.29.2, setPitch have issues when reaching the values 1.57-4.7 radians or 90-270 degrees. It is recommended to use setOrientation.
```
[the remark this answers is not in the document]
eGlint: I do not know about lep/jassdoc, but I did my research. I looked around lep/jassdoc git along with the forks and found nothing. 

Also, I just posted a detailed bug report about this issue in Hive.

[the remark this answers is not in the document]
Cokemonkey11: If it isn't already in lep/jassdoc then I would aks you to please contribute - https://github.com/lep/jassdoc/blob/master/effects.j

@lep fyi

Frotty:
nice, thanks 🍬

PR #88, opened 2018-06-16
Table: Add hasHandle and removeString methods.

crojewsk:
Following missing methods have been added:
- hasHandle
- removeString

Reorganized Table.wurst code to improve readability.


Frotty:
👍

PR #89, opened 2018-06-19
Added EnableUserControl()

Ouguiya:
Added function to enable / disable user control for the local player (similar to what ``SetUserControlForce`` does in blizzard.j)


Frotty:
Probably better change name to `setUserControl` and additionally add 2 functions in line of show/hide for showUnit, like enableUserControl() and disableUserControl().

Cokemonkey11:
LGTM

Frotty:
👍 Thanks

PR #91, opened 2018-07-15
Update Boolean.wurst and PrimitivesTests.wurst

Overkane:
Made shorter boolean functions and added tests for it


Frotty:
good stuff 👍

PR #94, opened 2018-07-21
[WIP] Various fixes in Networking/FileIO packages

SamuelMoriarty:


wurst/network/SyncSimple.wurst
```diff
@@ -117,8 +117,8 @@ public class SimpleSynchronizer
 
 	private unit dummy
 	private bitset syncedPlayers = emptyBitset()
-	// 4095 is 1111 1111 1111 in binary, one bit for each player slot
-	private static bitset allPlayers = bitset(4095)
+	// this sets allPlayers to a bit pattern of 1111..1111 where the total amount of 1's is equal to bj_MAX_PLAYER_SLOTS
+	private static bitset allPlayers = bitset(2 .pow(bj_MAX_PLAYER_SLOTS) - 1)
```
Cokemonkey11: will this be evaluated at compiletime?

wurst/file/PreloadIO.wurst
```diff
@@ -17,7 +17,7 @@ import ErrorHandling
 import StringUtils
 
 /** Maximum amount of packets per single readable file. */
-public constant PACKETS_PER_FILE = 16 	
+public constant PACKETS_PER_FILE = bj_MAX_PLAYER_SLOTS
```
Cokemonkey11: Can you explain how this is related to player slots?

wurst/network/SyncSimple.wurst
SamuelMoriarty: I'll double check this. Will wrap in `compiletime` if not.

wurst/file/PreloadIO.wurst
SamuelMoriarty: The current implementation relies on `SetPlayerName` to write/read packets to/from files. The previous limit was 16 'packets', equal to the number of players. Now that 1.29 supports 28 players, having it depend on the actual number of players as defined by the game better conveys the intent and is more reliable.

SamuelMoriarty:
Don't merge this yet. I'm currently working on my map and discovering more subtle bugs in the libraries. Will update when my own map is working stable.

wurst/network/SyncSimple.wurst
Cokemonkey11: I think it will work either way, I'm just interested

wurst/_handles/Unit.wurst
```diff
@@ -151,6 +151,9 @@ public function unit.getMoveSpeed() returns real
 public function unit.getName() returns string
 	return GetUnitName(this)
 
+public function unit.getProperName() returns string
```
Cokemonkey11: What's the behavior for non-heroes?

wurst/network/SyncSimple.wurst
SamuelMoriarty: Turns out it doesn't.

wurst/_handles/Unit.wurst
SamuelMoriarty: Simply returns null.

Cokemonkey11: Is it documented in lep/jassdoc? Can you document it here? Is it right for this to be an extension function of `unit`?

wurst/network/SyncSimple.wurst
Cokemonkey11: @Frotty might be interested to know that, thanks

wurst/_handles/Unit.wurst
SamuelMoriarty: 1) Jassdoc looks like it's hopelessly outdated. Someone needs to bring over all the new natives introduced into the latest patches there, but I am personally not up to the task. Just having one and not others would be weird.

2) Other hero-related functions are exposed as extension funcs on top of `unit`, so I don't see any consistency issues here.

Cokemonkey11: 1. I disagree about it being weird to have one but not others. If you don't have time to make a MR, can you at least raise an issue specifying that unit.getProperName returns null for non-hero?

2. Ok

Frotty: > Jassdoc looks like it's hopelessly outdated.

Why?
I don't think writing documentation about broken new natives, which get changed by pretty much any successive patch, has high priority.
Is there anything else outdated? I think jassdoc actually contains docs which cannot be found anywhere else. And unlike questionable pinned threads that only cover a small amount of natives in depth, this is an amazing, very structured and consistent overview.
And in its form it can be more easily used.

SamuelMoriarty:
Some comments on my last commit.

1) I added an even that gets executed after the LocalFiles check has been completed - this is necesarry because trying to use Persistable too early will result in an error.

2) Fixed a desync in LocalPlayers on this line
`network.getData().writeBoolean(LocalFiles.isEnabled())`

3) Extended MAX_PACKET_LENGTH to 512 characters because it actually seems to work as intended.

4) Turned bitset compilation in SyncSimple to compiletime.

wurst/_handles/Unit.wurst
SamuelMoriarty: I think the new natives are going to more or less stay in their current form from now on, apart from minor fixes to them, and it would be safe to include them and note any outstanding bugs, because some of them are quite useful.

Either way, I sent a PR to lep to update that particular function. I just thought it was one of the new natives - which, turns out, it isn't.

Frotty:
Nice 🍭

PR #97, opened 2018-07-28
Damage type fixes

Trokkin:
Moved from DamageModify PR.

### DIFF
--- a/wurst/_handles/Group.wurst
+++ b/wurst/_handles/Group.wurst
@@ -1,6 +1,7 @@
 package Group
 import NoWurst
 import Vectors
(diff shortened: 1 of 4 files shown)


Frotty:
The group stuff doesn't fix anything. it's a nice addition, but yeah, doesn't really need to be same PR.
Anyway, looks good, thanks a lot 🍆

Cokemonkey11:
>  I can't embrace the idea of PR being as small as one little commit when it is a part of some bigger change.

With some practice the work needed to manage your git dependency tree becomes less and more natural

PR #98, opened 2018-07-28
UnitId asset list, categorized by race and campaign

Trokkin:
All data is taken from unitUI.slk
Model and scale data is provided in comments to help recognize some units.


Frotty:
why for version 1.26a?
Looks neat, only thing which I already mentioned last time is that the names can become a bit long.
6 char-id 'ncap' vs `UnitIds_NightElfCampaign.corruptedAncientProtector` might be a bit annoying.
Might be fine though

wurst/_wurst/assets/UnitIds.wurst
```diff
@@ -0,0 +1,923 @@
...
+	static constant knight = 'hkni'
+	static constant steamTank = 'hmtt'
+	static constant rifleman = 'hrif'
+	static constant priest = 'hmpr'
+	static constant bloodElfDragonHawk = 'hdhw'
+	static constant altarofKings = 'halt'
+	static constant scoutTower = 'hwtw'
+	static constant gryphonAviary = 'hgra'
+	static constant castle = 'hcas'
+	static constant humanBarracks = 'hbar'
```
Cokemonkey11: maybe just Barracks since inside UnitIds_Human namespace?

wurst/_wurst/assets/UnitIds.wurst
```diff
@@ -0,0 +1,923 @@
...
+	static constant keep = 'hkee'
+	static constant blacksmith = 'hbla'
+	static constant humanLumberMill = 'hlum'
```
Cokemonkey11: Just lumberMill maybe

wurst/_wurst/assets/UnitIds.wurst
```diff
@@ -0,0 +1,923 @@
...
+	static constant workshop = 'harm'
+	static constant cannonTower = 'hctw'
+	static constant townHall = 'htow'
+	static constant guardTower = 'hgtw'
+	static constant farm = 'hhou'
+	static constant arcaneSanctum = 'hars'
+	static constant heroArchMage = 'Hamg'
+	static constant heroMountainKing = 'Hmkg'
+	static constant heroPaladin = 'Hpal'
+	static constant bloodElfSorceror = 'Hblm'
```
Cokemonkey11: heroBloodMage?

wurst/_wurst/assets/UnitIds.wurst
```diff
@@ -0,0 +1,923 @@
...
+	static constant waterElemental1 = 'hwat'
+	static constant waterElemental2 = 'hwt2'
+	static constant waterElemental3 = 'hwt3'
+	static constant militia = 'hmil'
+	static constant rocketTank = 'hrtt'
```
Cokemonkey11: ?

Trokkin: Maybe steamTankRocket

Frotty:
Not too sure about the scale comments etc. ? Will people really look into that?

I think 1 big group is better, but now the entries are not alphabetically ordered anymore, please restore that.

Trokkin:
As far as I used to, custom units should inherit properties of their model, like backswing point, shadow or missile firing point, thus the model is the key parameter when choosing base unit. Selection scale is also important and should correlate with the model scale and the model itself, so I focus on these too in hope that selection scale would be changed corresponding to scale change because I'm kinda tired of seeing inproper settings for units.

They never was alphabetically ordered, you know. At least for english: original data I was working with first was from 1.26a russian client, extracted by War3ObjectEditor which never worked with new versions, and there was alphabetical sort within categories.

Frotty:
🍬 Thanks

PR #102, opened 2018-08-07
Functions to work with triggeraction & triggercondition

Rhazarian:
These functions allow to use triggeraction & triggercondition in generics. It can be useful, e.g. if you manage triggers dynamically and want to remove actions and conditions to avoid leaks


Frotty:
Hi, thanks for your contribution.
Is this your last three PRs combined or what was the deal with closing them instantly?
Looks good to me.

Rhazarian:
Yes, it's those combined. The reason was that they weren't able to build separately because they depend on each other, actually pulling them in such a way was a mistake.

Frotty:
Good stuff 💎

PR #104, opened 2018-08-08
Fixed trigger API

Rhazarian:
Fixed existing functions to return values as original ones do & added missing ones


Frotty:
Good Stuff 👍

PR #105, opened 2018-08-08
Clamp method for Integers and Reals

Rhazarian:
Simple method which can be useful when you need to ensure that the number lies within a range


wurst/_handles/primitives/Real.wurst
```diff
@@ -93,3 +102,9 @@ public function real.lerp(real target, real alpha) returns real
 /** Checks if this real is between low and high value */
 public function real.isBetween(real low, real high) returns bool
 	return this >= low and this <= high
+	
+@test function test_clamp()
```
Frotty: should go into `PrimitiveTests`

Rhazarian:
Sorry for all the mess, the first errors were caused because of accident mixed indentation, but now I don't actually understand what's the thing with ambiguity here and how to solve it properly.

Frotty:
A value like `1` can represent either an int or a real. Therefore the call to the clamp function is ambiguous - both functions would work and wurst doesn't know which one to chose.
Fix that by casting to int or adding `.` for reals.

Frotty:
👍 
I think last step is to convert it to a ternary 1-liner similar to sign()

Rhazarian:
Done.
The issue with ambiguity was not obvious to me because I was used to the way how many other languages treat literals like `1` (as ints).

Frotty:
Nice 🍬 
Now please clean your fork :)

PR #106, opened 2018-08-14
Revert raising limit on preloader

SamuelMoriarty:
Contrary to my tests, WC3 can't read more than 223 characters from a player name, which broke the file IO. Reverting this to it's original value.


Frotty:
😅

PR #108, opened 2018-08-18
Better inlining in Printing.wurst

SamuelMoriarty:


Frotty:
Maybe add to description what the current problem is.
The correct solution might also be just to add/improve optimizations

Frotty:
Okay, the inlining rules have been improved. The example you posted now gets inlined.
Can you re-check on your project?

PR #109, opened 2018-08-29
Add matrices package

GetLocalPlayer:


Frotty:
Haha, I was just about to write a matrix lib myself this week.
Good stuff  👍 

Few things:
- Naming should be `mat2` and `mat3` like `vec` 
- Do you have the `rc` cell naming from somewhere? The one is now is `m00` etc. which is more common, but I might be wrong
- Some functions who could need it, are without hotdoc

I will add comments to the code and additional function suggestions later.

wurst/math/Matrices.wurst
```diff
@@ -90,43 +89,42 @@ public function matrix2.row(int id) returns vec2
         default
             return ZERO2
 
-public function matrix2.rotate(angle angl) returns matrix2
```
GetLocalPlayer: Removed this function which could be redundant or even useless since it represented common matrix multiplication.

GetLocalPlayer:
Quaternions package was added as well.

Frotty:
No, keep that to a separate PR.

PR #111, opened 2018-08-31
Matrices package

GetLocalPlayer:


wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+public function vec2.op_mult(mat2 m) returns vec2
+    return vec2(this.x*m.rc00 + this.y*m.rc10, this.x*m.rc01 + this.y*m.rc11)
+
+public function mat2.op_mult(mat2 m) returns mat2
+    return mat2(
+        this.rc00*m.rc00 + this.rc01*m.rc10, this.rc00*m.rc01 + this.rc01*m.rc11, 
+        this.rc10*m.rc00 + this.rc11*m.rc10, this.rc10*m.rc01 + this.rc11*m.rc11)
+
+/** Returns matrix' column by id as a vector or zero-vector if doesn't exist. */
+public function mat2.col(int id) returns vec2
```
SamuelMoriarty: Can you test how well this (and .row()) get inlined?
Consider using a temp-var maybe instead of separate return statements.

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+    0, 1)
+
+public let IDENTITY33 = mat3(
+    1, 0, 0,
+    0, 1, 0,
+    0, 0, 1)
+
+
+/* =========== 2X2 MATRIX =========== */
+/** Matrxi 2x2
```
SamuelMoriarty: Typo

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+        this.rc20 - scalar, this.rc21 - scalar, this.rc22 - scalar)
+
+public function mat3.op_mult(real scalar) returns mat3
+    return mat3(
+        this.rc00*scalar, this.rc01*scalar, this.rc02*scalar,
+        this.rc10*scalar, this.rc11*scalar, this.rc12*scalar,
+        this.rc20*scalar, this.rc21*scalar, this.rc22*scalar)
+
+/** Matrix multiplication is not commutative that means matrix*vect is not the same as vec*matrix. */
+public function real.op_mult(mat3 m) returns mat3
```
SamuelMoriarty: How is this comment relevant to scalar multiplication?

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+        this.rc10*m.rc00 + this.rc11*m.rc10 + this.rc12*m.rc20,
+        this.rc10*m.rc01 + this.rc11*m.rc11 + this.rc12*m.rc21,
+        this.rc10*m.rc02 + this.rc11*m.rc12 + this.rc12*m.rc22,
+
+        this.rc20*m.rc00 + this.rc21*m.rc10 + this.rc22*m.rc20,
+        this.rc20*m.rc01 + this.rc21*m.rc11 + this.rc22*m.rc21,
+        this.rc20*m.rc02 + this.rc21*m.rc12 + this.rc22*m.rc22)
+    
+/** Returns matrix' column by id as a vector or zero-vector if doesn't exist. */
+public function mat3.col(int id) returns vec3
```
SamuelMoriarty: Same consideration as with mat2 regarding inlining

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+
+/** Flips the matrix over its main diagonale.
+Result of transposing for rotation matrix is its inverse matrix. */
+public function mat3.transpose() returns mat3
+    return mat3(
+        this.rc00, this.rc10, this.rc20,
+        this.rc01, this.rc11, this.rc21,
+        this.rc02, this.rc12, this.rc22)
+
+/** Returns a zero-matrix in case there's no iverse matrix.*/
```
SamuelMoriarty: Doc comment should describe more than just it's corner case.
Also, consider adding a `.hasInverse()` function instead of returning a zero-matrix, or returning a result tuple like
`inverseresutl(boolean success, mat3 result)`.

Trying to check if this call succeeded will otherwise entail 9 comparison if you want to compare it against the zero-matrix.

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+            this.rc12*this.rc20 - this.rc10*this.rc22,
+            this.rc00*this.rc22 - this.rc20*this.rc02,
+            this.rc02*this.rc10 - this.rc00*this.rc12,
+
+            this.rc10*this.rc21 - this.rc11*this.rc20,
+            this.rc01*this.rc20 - this.rc00*this.rc21,
+            this.rc00*this.rc11 - this.rc01*this.rc10)
+    return ZERO33
+
+/** Extracts Euler-angles (Tait-Bryan) from rotation matrix. Order of result angles is X-Y-Z.*/
```
SamuelMoriarty: Why Tait-Bryan and not classical?

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+        roll = Atan2(sinZ, cosZ)
+    else
+        yaw = 0.
+        var cosZ = this.rc11
+        var sinZ = this.rc10
+        roll  = Atan2(sinZ, cosZ) 
+    return vec3(yaw, pitch, roll)
+
+public function mat3.toString() returns string
+    return "Matrix 3x3\n{0} {1} {2}\n{3} {4} {5}\n{6} {7} {8}".format(
```
SamuelMoriarty: Don't use .format(), it's very infefficient.

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+import NoWurst
+import Vectors
+import String
+
+
+/* Great thanks to euclideanspace.com for wide explanations. */
+
+
+public let ZERO22 = mat2(0, 0, 0, 0)
+public let ZERO33 = mat3(0, 0, 0, 0, 0, 0, 0, 0, 0)
```
SamuelMoriarty: Needs a better constant name.
Also, use `public constant` vs `public let`.

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+
+public let IDENTITY22 = mat2(
```
SamuelMoriarty: Ditto.

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+    1, 0,
+    0, 1)
+
+public let IDENTITY33 = mat3(
```
SamuelMoriarty: Ditto.

Frotty:
Yup, unit tests are mandatory.

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+        this.rc10 - scalar, this.rc11 - scalar)
+
+public function mat2.op_plus(real scalar) returns mat2
+    return mat2(
+        this.rc00 + scalar, this.rc01 + scalar,
+        this.rc10 + scalar, this.rc11 + scalar)
+    
+public function mat2.op_mult(real scalar) returns mat2
+    return mat2(
+        this.rc00*scalar, this.rc01*scalar,
```
Frotty: You leave spaces around `-` and `+` but not `*`. Pls fix for `*`

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+/** Flips the matrix over its main diagonale.
+Result of transposing for rotation matrix is its inverse matrix. */
+public function mat2.transpose() returns mat2
+    return mat2(
+        this.rc00, this.rc10,
+        this.rc01, this.rc11)
+
+/** Returns a zero-matrix in case there's no inverse matrix.*/
+public function mat2.inverse() returns mat2
+    var d = this.determinant()
```
Frotty: use `let` for constant locals.

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+    0, 1, 0,
+    0, 0, 1)
+
+
+/* =========== 2X2 MATRIX =========== */
+/** Matrxi 2x2
+00 01
+10 11
+*/
+public tuple mat2(real rc00, real rc01, real rc10, real rc11)
```
Frotty: Get rid of `rc`. Either use the mathematical convention `ij` or `m`

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,310 @@
...
+
+/** Creates 3x3 rotation matrix from axis and angle. */
+public function vec3.toMatrix(angle angl) returns mat3
+    var v = this.norm()
+    var x = v.x
+    var y = v.y
+    var z = v.z
+    var cos = angl.cos()
+    var sin = angl.sin()
+    return mat3(
```
Frotty: this is a prime example of why I don't like having an empty row `mat3(` followed by un-tabulated entries.
This to me just looks confusing/out of place.
I would suggest filling the first row and then properly aligning the cells, e.g.:

```
mat3(	1 + (1 - cos)*(x*x - 1), 	-z*sin+ (1 - cos)*x*y, 		y*sin+ (1 - cos)*x*z,
	z*sin+ (1 - cos)*x*y, 		1 + (1 - cos)*(y*y - 1), 	-x*sin+ (1 - cos)*y*z,
	-y*sin+ (1 - cos)*x*z, 		x*sin+ (1 - cos)*y*z, 		1 + (1 - cos)*(z*z - 1))
```
This applies to all of these statements above, especially mat2 with short params.

Also these calcs are overly big. You should e.g. save 1 - cos in a var

wurst/math/Matrices.wurst
GetLocalPlayer: As far as I know an operation with high priority doesn't require space around it. In most languages. That means I should leave one space around ``-`` and ``+`` but shouldn't do that around ``*`` and ``/``. In this case you ask me to fix it for ``*`` but not for ``/``. Is there any wurst-style doc I need to learn?

wurst/math/Matrices.wurst
GetLocalPlayer: For such kind of test we have to calc determinant and once more in ``.inverse`` if it's not 0. More operations that's undesirable.  I think tuple-result is the best solution thanks for hint.

wurst/math/Matrices.wurst
GetLocalPlayer: Funny, thanks.

wurst/math/Matrices.wurst
GetLocalPlayer: Didn't think about inlining at all. You're right, I'll rewrite it with a temporary variable.

wurst/math/Matrices.wurst
GetLocalPlayer: Is performance above readability in such debug functions?

wurst/math/Matrices.wurst
GetLocalPlayer: Thanks

GetLocalPlayer:
Most matrices tests were added. Cannot produce Mat2Euler tests
* `rc` replaced with `m`
* `.inverse` now returns `inverseresult22` and `inverseresult33` tuples
* Such strings as 
```
    return mat2(
                this.m00*scalar, this.m01*scalar,
                this.m10*scalar, this.m11*scalar)
```
rewritten as
```
    return mat2(this.m00*scalar,    this.m01*scalar,
                this.m10*scalar,    this.m11*scalar)
```


> FAILED with exception: You encountered a bug in the interpreter: java.lang.Error: The native <Asin> has not been implemented for compiletime!

Frotty:
> You encountered a bug in the interpreter: java.lang.Error: The native Asin has not been implemented for compiletime!

I will add it.

Regarding spacing around operators, intellij simply spaces every, and I don't see any point in having some unintuitive rule for this. Just space everything consistently. if you want to clarify parts of a term, use brackets or local vars.

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,418 @@
...
+    let rotation = axis.toMatrix(a)
+    print("Rotation around {0} by {1} degrees is rotation {2}\n".format(axis.toString(), a.degrees().toString(), rotation.toString()))
+    // To Euler angles
+    // Cannot produce this tests bcz Asin is not implemented yet.
+    let euler = rotation.toEuler()
+    print("Found rotation matrix to Euler angles {0}".format(euler.toString()))
+    
+
+@test
+function test()
```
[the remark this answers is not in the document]
GetLocalPlayer: What else should Iknow to not rewrite it again again and again. How many instructions do you have and why I can't see the whole list?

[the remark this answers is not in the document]
Frotty: You could check out existing packages before doing something very different.

There is no list _yet_, but I also thought about it being useful in the future. Should probably start a code conventions doc.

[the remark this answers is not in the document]
GetLocalPlayer: And I did. For example, `Vectros` package contains `*` with space around it and without or `Angles` package. But  have to.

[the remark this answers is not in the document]
Frotty: I'm talking about the tests. Obviously many packages are old, made or modified by different people and, resulting in different styles. Especially before we had proper PRs and reviews.
New additions however, should follow the convention.
Soon the formatting should be standardized, if @karlek finishes https://github.com/wurstscript/WurstScript/pull/620 and then we can easily format the stdlib consistently.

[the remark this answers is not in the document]
GetLocalPlayer: > New additions however, should follow the convention. 

Which doesn't exist.
Dunno how many surprises are waiting for me. Maybe later.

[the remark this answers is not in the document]
Frotty: I'm sorry to hear you are disappointed by the pull request process, but I don't think it's anything out of the ordinary.
It can easily take weeks before a pull request is accepted, on any project.Just check other PRs on our projects and other repositories.

I really don't understand why you get so frustrated. We appreciate your contribution, just take a little time to do improvements. This isn't critical.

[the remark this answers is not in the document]
GetLocalPlayer: May be because I hardly distinguish speech styles but not, I'm not frustrated.
I'm not interested in WC3 much enough to keep going hint2hint in a black room so I decided to  close this until there's the convention and correct list of requirements. As I said, maybe latter.

GetLocalPlayer:
How reliable is [this](https://wurstlang.org/manual.html#coding-conventions)?

Frotty:
Still subject to change

GetLocalPlayer:
Can tests be done in separate files for each matrix type? Can custom `mat2.assertEquals` function be used with `testFails` in it instead of `real.assertEquals`. Or it can be only those ones that are defined in `Wurstunit`?

GetLocalPlayer:
Tests were united. ``MatricesTests`` pacgake was added.

wurst/math/MatricesTests.wurst
```diff
@@ -0,0 +1,297 @@
...
+        testFail(msg.format("m12", expected.m12.toString(), this.m12.toString(), delta.toString()))
+
+    if compare.m20 > delta
+        testFail(msg.format("m20", expected.m20.toString(), this.m20.toString(), delta.toString()))
+    if compare.m21 > delta
+        testFail(msg.format("m21", expected.m21.toString(), this.m21.toString(), delta.toString()))
+    if compare.m22 > delta
+        testFail(msg.format("m22", expected.m22.toString(), this.m22.toString(), delta.toString()))
+
+function vec3.assertEquals(vec3 expected)
```
Frotty: should go into the vector package I assume

Frotty:
Looking good 👍

wurst/math/MatricesTests.wurst
GetLocalPlayer: Hmm... Would it better to move mat3.assertEquals from MatricesTests into Matrices as well? And do the same for each similar function in the future.

GetLocalPlayer:
Jesus, what a mess I've done. Should I do anything with it? I'm sorry to ask, I'm quite newbie in Git.

Frotty:
Many commits != mess. I will squash them all into one commit on merging anyway.
You could start using git on your machine, instead of using the web interface, but doesn't really matter.

GetLocalPlayer:
It is done.

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,352 @@
...
+/** Returns matrix' column by id as a vector or zero-vector if doesn't exist. */
+public function mat2.col(int id) returns vec2
+    vec2 result
+    switch id
+        case 0
+            result = vec2(this.m00, this.m10)
+        case 1
+            result = vec2(this.m01, this.m11)
+        default
+            result = ZERO2
```
SamuelMoriarty: IMO, this is an invalid use of the API and should result in an error instead of a zero-value being returned.

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,352 @@
...
+/** Returns matrix' row by id as a vector or zero-vector if doesn't exist. */
+public function mat2.row(int id) returns vec2
+    vec2 result
+    switch id
+        case 0
+            result = vec2(this.m00, this.m01)
+        case 1
+            result = vec2(this.m10, this.m11)
+        default
+            result = ZERO2
```
SamuelMoriarty: Ditto.

wurst/math/Matrices.wurst
```diff
@@ -0,0 +1,352 @@
...
+    vec3 result
+    switch id
+        case 0
+            result = vec3(this.m00, this.m01, this.m02)
+        case 1
+            result = vec3(this.m10, this.m11, this.m12)
+        case 2
+            result = vec3(this.m20, this.m21, this.m22)
+        default
+            result = ZERO3
```
SamuelMoriarty: Ditto.

wurst/math/Matrices.wurst
GetLocalPlayer: Such functions (as .inverse ones) should return `null` if result can't be obtained. Since we use tuples which can't take `null` and since WC3 doesn't support exceptions how can we do it here? `inversereulst` was a good solution for `.inverse` functions but such approach would be surplus here.
I completely agree with you, but I have no idea.

SamuelMoriarty: Just call `error()`.

I don't see any use-case where using this function with `n` out of bounds would result in any kind of 'desirable' result.

When calculating an inverse, you don't know ahead of time if there's one or not, so we return a result tuple.

When getting a row, however, you *do* know if there's a row or not, so passing in an invalid value is a mis-use of the API.

Hence, just call `error()` specifying the error.

I might submit a PR some time later with some kind of exception-handling mechanism for use instead of error(), since all the infrastructure is there already, but for now, just use `error()`.

GetLocalPlayer: > When calculating an inverse, you don't know ahead of time if there's one or not, so we return a result tuple.

We can find it from pre-calculation of the determinant and then compare obtained result with 0. But it gives us many extra calculations. Moreover, since the role of this package, obviously, is rotations we always know there's the inverse matrix and we can obtain it through `.transpose` without any calculations.

Meh, just lyric.

I agree and will add to the code `error` calls. Obviously is much better to know where the error from.

wurst/math/Vectors.wurst
```diff
@@ -3,7 +3,9 @@ import NoWurst
 import Wurstunit
 import public Real
 import public Angle
+import String
 import initlater Printing
+import initlater Matrices
```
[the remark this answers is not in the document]
GetLocalPlayer: Fixed.

Frotty:
For all multiplication operations you should specify in hotdoc, whether you are using pre- or post-multiplication.

Also, you offer creation of rotation and scaling matrix, so please add translation matrix as well. Rename "rotation matrix" related funcs accordingly. You should also add a `toRotation(angle val)` that takes just an angle.

GetLocalPlayer:
> For all multiplication operations you should specify in hotdoc, whether you are using pre- or post-multiplication.

This is not clear for me.

> Also, you offer creation of rotation and scaling matrix, so please add translation matrix as well. 

Translation matrix for vec2 is 3x3 matrix (with implicit conversion vec2 to vec3). Translation matrix for vec3 is 4x4 matrix. First can be done easily. Do you require to add mat4 in the package?

> You should also add a mat3.toRotation(angle val) that takes just an angle.

What should such function do? Any 3D rotation can be represented as axis-angle but just an angle?

GetLocalPlayer:
> But i guess it's fine due to form of extension functions implying it already.

Well, the non commutative functions are `op_mult` where pre/post-multiplication is described by an expression itself `vec * mat` or`mat * vec`.

> vec2 to mat3 trans should suffice for now

Will it be confusing if I overload `op_mult` for `vec2` in case `mat3 * vec2 = vec2`? 

> unless you're very keen on making a mat4 tuple reality :)

Nah. Wc shows a strong performance fall in work with several primitives (square/box/pyramid) through 3x3 matrices.

> It set the matrix to a rotation matrix that will rotate any vector in counter-clockwise direction around the z-axis, meaning 0,0,1 for the 3rd row

Oh, a short alternative to `angle.toRotation(vec3 axis)' where `axis` is `vec3(0, 0, 1)`, right?

GetLocalPlayer:
Done.
Instead of `.toRotation()`, `.toRotationX()`, `.toRotationY()`, `.toRotationZ()` were added.

Frotty:
IDENTITY22 and IDENTITY33 are still indented in old style.
Other than that looking good.

GetLocalPlayer:
Oh, yes. How could I miss it.

Frotty:
@SamuelMoriarty do you approve the changed?
LGTM now 👍

PR #112, opened 2018-08-31
Quaternions package

GetLocalPlayer:
Depends on #111


wurst/math/Quaternions.wurst
```diff
@@ -0,0 +1,107 @@
...
+        return q
+    // choosing the longest way
+    return dot < 0 ? this.slerp(q, p) : this.slerp(quat(-q.x, -q.y, -q.z, -q.w), p)
+    
+public function quat.toString() returns string
+    return "Quaternion [X {0}, Y {1}, Z {2}, W {3}]".format(this.x.toString(), this.y.toString(), this.z.toString(), this.w.toString())
+
+/** Extracts Euler-angles (Tait-Bryan) from a quaternion. Order of result angles is X-Y-Z.
+There's no unic way for quaternions thus we have to use quat-matrix conversion first. */
+public function quat.toEuler() returns vec3
```
SamuelMoriarty: Using vec3 for representing a euler angle is a bit misleading here. While technically the representation is the same, semantically these are very different mathematical concepts, and most operations applied to vectors do not make sense in the context of euler angles.

Unless what you mean to imply here by Euler angle is actually a directional unit-vector... Which afaik is not an Euler angle at all.

A separate tuple for euler angles (and maybe even a separate package with functions to operate on them) would be ideal IMO.

SamuelMoriarty: I just realized you mentioned Tait-Bryan angles in there.
Is there a particular reason for using those instead of classical Euler angles? As far as I know, the typical usage of Euler angles in games is the classical Euler angle, i.e. Pitch, Yaw, Roll, and not Yaw, Pitch, Roll

Frotty:
- Use `let` if locals aren't modified.
- Same thing about `quat(`
- Add unit tests

GetLocalPlayer:
`QuaternionsTests` package was added.


> - Use let if locals aren't modified.
> - Same thing about quat(
> - Add unit tests

Done.

GetLocalPlayer:
`.assertEquals` function for quaternion was moved into `Quaternions` package.
`.assertEquals` functions for vec3 and mat3 were deleted according to dependence on #111 
Added functions `angle.toQuat` and `quat.toMatrix`. Added test functions according the new ones.

GetLocalPlayer:
Requirements have been met. I believe I won't add anything new.

wurst/math/Quaternions.wurst
```diff
@@ -0,0 +1,215 @@
...
+q - targeted quaternion. p - progress from 0 to 1. */
+public function quat.lerp(quat q, real p) returns quat
+    quat result
+    if this.dot(q).abs() < 0.002
+        result = q
+    else
+        result = this * (1 - p) + q * p
+    return result
+
+/** Linear intERPolation.
```
SamuelMoriarty: Is this a joke about ERP? lol

Frotty: yea but why the casing

GetLocalPlayer: Because I cannot show it otherwise?

Frotty:
Good stuff so far, a few points:
* you return raw angle values instead of `angle` tuples from certain functions. Might not be easily solvable, but at least define whether it's in rad or deg.
* Add a gimbal lock detection e.g. `getGimbalPole()`
* Add `exp` func see https://en.wikipedia.org/wiki/Quaternion#Exponential,_logarithm,_and_power_functions

wurst/math/Quaternions.wurst
Frotty: show what? the word is "interpolation", you write "intERPolation", you link some slerp/nlerp stuff which seems totally unrelated?

GetLocalPlayer:
> you return raw angle values instead of angle tuples from certain functions. Might not be easily solvable, but at least define whether it's in rad or deg.

But there are no functions that return an angle. There are only 2 functions in the package which return `real`. First is `.length` that is not related to angles. Second is `.dot` but dot product of two quaternions is not the angle, it's cos of half the angle between two rotations hence it will be incorrect to keep dot product in such container as `angle`.

Frotty:
toEuler..

GetLocalPlayer:
Which is one function and it returns vec3 but ok I'll add information in hotdoc.

Frotty:
yes, and the x,y,z values of that vector are raw angles..

Frotty:
Yea, good stuff. I would probably name the File just `Quaternion` since it doesn't hold multiple datastructures like vecturs/matrices.

Frotty:
> Took thought about adding .exp

`Quaternion.exp(real alpha) returns Quaternion`

Doesn't require logarithm, just `pow` which exists.

As for the others, there of course are logarithm libraries in Jass, but you're right, nothing native, so I think `exp` should suffice. When a PR for log is added, it can always be added later.

GetLocalPlayer:
> there of course are logarithm libraries in Jass

Which may be not the best solution just because Jass is slow. May be it would be better to ask Blizzard to add `log` native.
Anyway, I don't think there's a case in Wc3 that would require the whole range of quaternion functions.

> Doesn't require logarithm, just pow which exists. it just calculates quat^alpha, where alpha is a real.

Yep, that's done. So, is there any requirement left?

Frotty:
> May be it would be better to ask Blizzard to add log native.

Good luck with that.

> Yep, that's done. So, is there any requirement left?

Implementation could be improved I think, but other than that I think nope.

Frotty:
Very nice 🍬

PR #116, opened 2018-09-12
Bitwise library

SamuelMoriarty:


Frotty:
Some comments for all 3 PRs:
- I think tests and other top level declarations shouldn't be mixed. I expect there to only be tests after the first. Easiest solution I guess is to extract them into *Tests packages.
- you usage `integer`, `int`, `real` etc. Prefer type inference and stick to `int` by convention.
- The arrays could probably be constant. Any reason against that?
- Use the extension functions `.assertEquals()` instead of `assertTrue(a == b)` . It will have better feedback upon failure.

Specific to this:
- Please add docs to functions about their pendants which they roughly represent (>> -> shiftR whatev) and give some small examples on how to translate such code.
- use better variable names where it makes sense

SamuelMoriarty:
Moved bitwise init data into `BitwiseInit.wurst`
Moved bitwise tests into `BitwiseTests.wurst`

All core logic remains in `Bitwise.wurst`

Now using `.assertEquals` vs `assertTrue`
Now using `var` and `let` where applicable.
Extended docs for extension functions.

SamuelMoriarty:
> The arrays could probably be constant. Any reason against that?

Probably, but I'm a bit too lazy to go and reformat them into constants. I think they are fine as it is.

wurst/bitwise/Bitwise.wurst
```diff
@@ -0,0 +1,316 @@
...
+/** 32-bit SHIFTR operator **/
+function bwShiftr32(int num, int shift) returns int
+    var value = num
+    var isNegative = 0
+    if value < 0
+        value = -2147483648 + value
+        isNegative = 1
+    return isNegative * powShift[31-shift] + value div powShift[shift]
+
+// function bwRotl32(int val, int shift) returns int
```
Frotty: what's this commented stuff?

SamuelMoriarty: bitwise rotation
it was in the library that I used as a reference, but it wasn't correctly working for negative numbers, so I commented it out - seeing as it wasn't really all that necessary anyway
I'm not sure if I can fix it

Frotty:
Sweet 🍰

PR #118, opened 2018-09-12
New FileIO library

SamuelMoriarty:
Depends on #117


wurst/file/Encryptor.wurst
```diff
@@ -0,0 +1,13 @@
+package Encryptor
```
Frotty: Shouldn't this be more generic, like Processor/Pipeline/(De)serializer ?

SamuelMoriarty: FWIW, I think this could be called (Persistable)Middleware.

PR #123, opened 2018-09-26
Update LinkedList.wurst raises warnings

GetLocalPlayer:
Cannot use `LinkedList` with handles by itself
screenshot_4
It requires to import `TypeCasting` package but then we get warnings
screenshot_5
Can be resolved through public import of `TypeCasting` withing `LinkedList`.

### DIFF
--- a/wurst/data/LinkedList.wurst
+++ b/wurst/data/LinkedList.wurst
@@ -1,6 +1,6 @@
 package LinkedList
 import NoWurst
-import TypeCasting
+import public TypeCasting
 import Integer
 import String
 import ClosureForGroups


Frotty:
> Cannot use LinkedList with handles by itself
> It requires to import TypeCasting package but then we get warnings

As intended. [There is a ticket for the warning](https://github.com/wurstscript/WurstScript/issues/381).
Adding this as public import to LinkedList doesn't make sense since this applies to all generic classes.
If it should be public, it should just be in `Wurst,wurst`.
IIRC the reason against making it a public import was the hackish nature of the package and the potential of breaking with upcoming patches.

PR #124, opened 2018-09-27
Add vec.isInPolygon

GetLocalPlayer:
Redused calcs in `vec2.isInTriangle` 
I wouldn't like to copy the code from `vec2.isInPolygon` into `vec3.isInPolygon2d`.


GetLocalPlayer:
I'm likely to write Raycast package then we'll see what's left to implement.

PR #125, opened 2018-10-04
Raycast.wurst

GetLocalPlayer:
Contains

For 2D:
- [x] point projection onto ray
- [x] ray X line
- [x] ray X ray
- [x] ray X segment
- [x] ray X circle 

For 3D:
- [x] point projection onto ray
- [x] ray X plane
- [x] ray X disc
- [x] ray X sphere
- [x] ray X triangle

Of course, there are lots of extra things that can be implemented but they can be surplus in Wc3 for stdlib.

### DIFF
--- a/wurst/math/Raycast.wurst
+++ b/wurst/math/Raycast.wurst
@@ -3,24 +3,45 @@ import NoWurst
 import Vectors
 import Wurstunit
 
+/* The package defines such geometry as 'ray' for 2D and 3D separately.
+Both can find the intersection point with another geometry objects.
+
+Most 2D functions return result2d(success, point, distance) tuple which contains
+information about intersection where
+'success' is 'true' if the ray intersects the given object
+'point' is the intersection point or zero-vector
+'distance' is the distance between the ray's origin and the intersection point
+
+A ray can intersect a convex object (e.g. a circle) zero, one or two times at once.
+In this case the function returns doubleresult2d tuple that contains a couple of result2d
+tuples.
+
+Functions that find 3D intersection work similarly and return result3d tuple.
+result3d(success, point, distance, backface) also contains 'backface' boolean value that's
+true in case the object is faced away from  the ray's origin. Functions which find the
+intersection point with a convex object (e.g. a sphere) return doubleresult3d tuple.
+
+Also you can project a vector onto a ray to obtain the point on the ray that's nearest
+to the vector. */
+
 constant FAIL_2D = result2d(false, ZERO2, 0)
 constant FAIL_3D = result3d(false, ZERO3, 0, false)
 constant EPSILON = 0.0001
 
 /** An infinite 2D ray
 origin - the point which the ray comes from
-direction - unit directional vector */
+direction - directional vector */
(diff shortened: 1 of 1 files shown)


Frotty:
The 3d functions look a bit messy, but other than that good stuff, though on the brim for stdlib.
You should recheck doc for typos "if it behind" and add some package doc where you explain the result tuple and perhaps a small wc3 use case.
Sry for taking some time to check up.

GetLocalPlayer:
> add some package doc where you explain that functions return the result tuple 

Can I add the hotdoc before `package Raycast` at the top of the file?

> though on the brim for stdlib.

There are several functions that I haven't dared to add

For 2D:
- ray X rectangle
- ray X convex polygon

For 3D:
- ray X box
- ray X cylinder (cone)

Do these fit stdlib? Are these overkill?

Frotty:
> Can I add the hotdoc above package Raycast at the top of the file?

I prefer it after the package and import directives. You can use normal /* */

> Do these fit stdlib? Are these overkill?

Probably overkill, also depends on size of course.

GetLocalPlayer:
There's some [map example](https://github.com/GetLocalPlayer/RaycastExample) if it could be useful. Dunno why.

wurst/math/Raycast.wurst
```diff
@@ -0,0 +1,483 @@
...
+backface - `true` if the surface at the intersection point is faced away from the ray's origin  */
+public tuple result3d(bool success, vec3 point, real distance, bool backface)
+
+/** Double 3D intersection result.
+'first' is nearest to the ray's origin. */
+public tuple doubleresult3d(result3d first, result3d second)
+
+/* =============== 2D =============== */
+
+function vec2.cross(vec2 v) returns real
```
Frotty: why is this here? Belongs into vector

wurst/math/Raycast.wurst
```diff
@@ -0,0 +1,483 @@
+package Raycast
+import NoWurst
+import Vectors
+import Wurstunit
+
+/* The package defines such geometry as 'ray' for 2D and 3D separately.
```
Frotty: You should add a small code snippet of one of the raycasts in practical use, showing how to handle the returned tuple correctly.

wurst/math/Raycast.wurst
GetLocalPlayer: Because there's no cross product of two 2d vectors. This is a simplification for several functions within the package.

GetLocalPlayer:
ain't sure how it can be incorrect. Did you mean that

Frotty:
Yea that's good. Last thing I'm thinking about is whether "success" is the correct wording, since nothing really succeeds here. `if ray.intersects` sounds better than `if ray.success` to me, so that might be better?

GetLocalPlayer:
I think you're right. `success` would be better for functions like `.intesectSomething` but not for raycasting where any result is basically succeed.

Frotty:
I think except for spaces-tabs issue this looks good now 👍

GetLocalPlayer:
Whitspaces replaced with tabs.

PR #128, opened 2018-10-14
Update Unit.wurst

GetLocalPlayer:


wurst/_handles/Unit.wurst
```diff
@@ -606,3 +606,15 @@ public function unit.getLocalZ() returns real
 
 public function unit.isIllusion() returns bool
 	return IsUnitIllusion(this)
+
+public function unit.toGroup(group g)
```
Frotty: These unit. functions seem to be pointless duplicates of existing functionality.
Since you need access to both unit and group, you should just use container.xxx(element)

Frotty:
sry, i only ment the group functions. `IncUnitAbilityLevel` doesn't seem to be wrapped yet, so that's good :)

GetLocalPlayer:
Oh, was a bit confused :grimacing:

PR #132, opened 2018-10-31
Fogmodifier missing functions

GetLocalPlayer:


Frotty:
nice, but i just noticed that you used spaces instead of tabs in all your contributions, please change to tabs. You can have this done automatically via `.editorconfig`

GetLocalPlayer:
Whitspaces replaced with tabs.

Frotty:
Does a fog modifier need to be stopped before destroying?

GetLocalPlayer:
Nah, it stops by itself.

```
let fog = createVision(Player(0), gg_rct_Test, true)
    ..start()

init
    CreateTrigger()..registerPlayerChatEvent(Player(0), "destroy", true)
    ..addAction() ->
        fog..destr()
```

ezgif-6-3d92d900eb1b

PR #135, opened 2018-11-11
Weather package

GetLocalPlayer:


GetLocalPlayer:
Instead of WeatherEffect asset

Frotty:
> Instead of WeatherEffect asset

Why? You should add descriptions to your PRs.
In this case I don't see the benefit and it's a breaking change if it were to replace WeatherEffects.

GetLocalPlayer:
Since weathereffect is a handle it requires proper placement. Furthermore that leads us to the necessity of proper function naming to avoid native functions usage that's beyond of the existing package.

Frotty:
The extension functions are fine, the replacement of WeatherEffects not.
You should also add this package as import to the `_Handles` package.

GetLocalPlayer:
Is there any sense to make users import "WeatherEffects"? It'll confuse them.

Frotty:
you can import public the package of course. But we don't want duplicated code nor breaking changes.
Can one make custom weather effects? If not it might be also useful to add additional safety

Frotty:
Sweet 🍭

PR #136, opened 2018-11-13
Added common extensions.

jonathanmclaus:


wurst/_handles/Unit.wurst
```diff
@@ -2,6 +2,7 @@ package Unit
 import NoWurst
 import public Vectors
 import Item
+import initlater HashMap
```
Frotty: we don't want initlater imports in stdlib. resolve it.

Frotty:
Thanks for the PR 😃 
Adding missing ext functions is always nice 👍 
I'm not a big fan of `getInventoryItemCountMap` though. For once the initlater import and it returns an object that users who are new or don't pay attention might not destroy as they should.
I guess this is for item recipes? Another idea might be just a vararg function that takes itemids and checks if the unit has all of them, omitting the need for hashmap.
Also, `getLoc` should probably come with @deprecated annotation so no1 uses it 🤔

wurst/_handles/Unit.wurst
```diff
@@ -627,6 +641,10 @@ public function unit.endAbilityCooldown(int abilCode)
 public function unit.getAbilityManaCost(int abilId, int level) returns int
 	return BlzGetUnitAbilityManaCost(this, abilId, level)
 
+	
+public @deprecated("Use getPos instead.") function unit.getLoc() returns location
```
Frotty: annotation should be above the func or at least before any other keyword.

Frotty:
Sry, I thought I had merged this already. Wanted to use that function myself now 😅 
Thanks again for the contribution 👍

PR #138, opened 2018-11-15
Blizzard changed events id

yefq:
These player unit events id was changed to 272~280 in the new vision.
And added player mouse event.

PR #152, opened 2018-12-01
Mouse events update

anufis:
Mouse events update


Frotty:
Nice, makes #138 obsolete 👍

PR #154, opened 2018-12-01
Update Effect.wurst

GetLocalPlayer:
New natives.
Currently I do not see any result from *SubAnimation functions.


Frotty:
pls merge master into this so travis re-checks

GetLocalPlayer:
Did it succeed?

Frotty:
No it can't because it requires new compiler version for updated common.j

PR #156, opened 2018-12-03
Update StringUtils.wurst

GetLocalPlayer:
Prevent `bezier_3` warnings.


Frotty:
i guess that whole function needs an update witht he widescreen support having shifted seemingly everything, but nice 👍

PR #143, opened 2018-11-18
add dialog handling for closure timers

Cokemonkey11:

PR #157, opened 2018-12-03
Update ClosureTimers.wurst

GetLocalPlayer:
Currently I've got some trouble with these closures. Sometimes I need to destroy an early prepared closure callback before its starting that raises an error "Trying to release a null timer".

I'm unsure, can I do this if another PR is existing #143? I see a bit more problems which should be solved.


Frotty:
Why close?

GetLocalPlayer:
Because I still didn't receive any answer.

Frotty:
Feel free to highlight the PR again with a comment including my name then.
Often I read the emails when I can't react to them, and then forget about it later.
No bad intentions.

PR #161, opened 2018-12-16
add hashmap size

PhoenixZeng:


wurst/data/HashMap.wurst
```diff
@@ -29,6 +33,17 @@ public class HashMap<K,V> extends Table
 		remove(key)
 		return result
 
+	/** Remove all data from this map */
+	override function flush()
+		size = 0
+		super.flush()
+
+	function size()returns int
```
Frotty: add space after )

wurst/data/HashMap.wurst
```diff
@@ -29,6 +33,17 @@ public class HashMap<K,V> extends Table
...
+		return size
+
+	ondestroy
```
Frotty: seems pointless? Shouldn't call anything on a destroyed instance.

wurst/data/HashMap.wurst
```diff
@@ -104,9 +119,11 @@ public class IterableMap<K,V> extends HashMap<K, V>
 
 
 	/** Returns the length of this IterableMap **/
-	function size() returns int
+	override function size() returns int
 		return keys.size()
 
+	function mapSize() returns int
```
Frotty: why is this here? size() and mapSize() should always be the same, right?

PhoenixZeng: yes. and i use it in the hashmaptest.wurst to check the size . it doesn't make any sense. Do I need to get rid of it?

PhoenixZeng:
`Few comments. I think there was an argument against adding size like this, because it adds "has" overhead to "put" and "remove". Idk, probably still fine tho.`
maybe i should make a sub class

wurst/data/HashMap.wurst
Frotty: Yes

wurst/data/HashMap.wurst
Frotty: Accessing values of destroyed instances is invalid

PhoenixZeng: i already maked a new commit

Frotty:
I'm not sure, I don't like having subclass just for size, but especially with typecasting the extra overhead could become important. Why not use `IterableMap` ? 🤔

PhoenixZeng:
because the `IterableMap ` has a hashlist . That's more overhead then Sizedmap. 

for sizedlist:
The point is to know how many instances a class currently has. (Because map mapping some jass-object to class is common used in wurst.)
What can this bring?
1: For classes that do not need to destroy instances, only integer loops are needed to traverse instances, and linked list modules are not needed (the iteration overhead of linked list modules is higher).
2: easy for instances leaks check (especially for 8191 of 1.27)
3: If you use integers as keys, this is also the a implementation of `HashList` (I don't like `LinkedList` because it needs to create `LLEntry`)

Frotty:
what is "sizedlist" ?

> For classes that do not need to destroy instances, only integer loops are needed to traverse instances, and linked list modules are not needed (the iteration overhead of linked list modules is higher)

unless you have nested iterations, you can save a LinkedListModule iterator and there won't be any allocation overhead.

> easy for instances leaks check (especially for 8191 of 1.27)

What does this have to do with maps? add a size var to the class, or make a module that does it for convenience.

> (I don't like LinkedList because it needs to create LLEntry)

Do you have an actual scenario where this is a problem?
LinkedLists perform best for most usecases. We don't want to split list structures up into too many different types (e.g. we had ArrayList, which might get a return) mostly because if they would all implement an interface, they would share id space. Otherwise you need to duplicate a lot of code for both types of list.

wurst/data/HashMap.wurst
```diff
@@ -29,6 +29,28 @@ public class HashMap<K,V> extends Table
 		remove(key)
 		return result
 
+public class SizedMap<K,V> extends HashMap<K,V>
```
Frotty: SizedMap kinda gives the illusion that the Map is constrained by a size, rather than just counting imo. @cokemonkey11 ?

Cokemonkey11: I guess a size-constrained one might be called "SmallMap" or something. SizedMap feels to me like `Map<> implements Sized` so name feels ok

I don't see the point though in extending hashmap. Why not just add size to hashmap?

PhoenixZeng:
`you can save a LinkedListModule iterator` 
yes .But that code would be more complicated than that. `for loop` is easy to use. and  Although nested iterations are used sparingly, they may also be used.
------------------------------------
idk. differert man have the differert uses for wurst. A subclass have only a size maybe not a good idea for the STD. This is a personal  class . I just need to add it to my library.

If there is no objection, we can close it.

wurst/data/HashMap.wurst
Frotty: guess the only argument is the introduced overhead.
If you think it doesn't matter, we should just add it to HashMap.

PhoenixZeng: `guess the only argument is the introduced overhead.`
i think add more overhead is unnecessary (with only a feature )

Halithor: Instead of adding this to the HashMap, why not just use the `IterableMap`? It already has the `size` method with minimal overhead (memory increase based on the number of keys). 

If we still want size for the a standard HashMap, I would prefer it goes into the standard HashMap. `size` methods are not uncommon methods for HashTable data structures and it would make sense for the Wurst implementation to have one.

PhoenixZeng: because the `IterableMap` has a inner `hashlist` .
if u think `adding this to the HashMap` is good. i can do it

Frotty:
just add it to hashmap then.

Frotty:
can you fix the build?

PhoenixZeng:
it fixd

wurst/data/HashMap.wurst
```diff
@@ -7,12 +7,16 @@ import public Table
...
+	private int size = 0
+	
 	/** Whether a value exists under the given key or not */
 	function has(K key) returns boolean
 		return hasInt(key castTo int)
 
 	/** Saves the given value under the given key */
 	function put(K key, V value)
+		if not has(key)
+			size ++
```
Frotty: Do not put spaces around unary operators

wurst/data/HashMap.wurst
```diff
@@ -7,12 +7,16 @@ import public Table
 /** Generic Table Wrapper */
 public class HashMap<K,V> extends Table
 
+	private int size = 0
```
Frotty: should probably be protected

wurst/data/HashMap.wurst
```diff
@@ -29,6 +35,16 @@ public class HashMap<K,V> extends Table
...
 		return result
 
+	function size() returns int
+		return size
+
+	override function flush()
+		size = 0
+		super.flush()
+		
+	ondestroy
```
Frotty: pointless, this already happens in Table

Frotty:
few more things

PR #168, opened 2019-01-05
Add ClosureTimers functions as extension methods of `timer`

theQuazz:
This allows for flexibility to use ClosurerTimers while keeping
a reference to the timer itself. This is especially important when
dealing with `timerdialog`.


Frotty:
Nice 🍨

PR #183, opened 2019-02-11
Layer -> ImageLayer

WareditorGit:
Layer is too generic, it should be left available.


Frotty:
I can see the issue, however I am not keen on merging this as long as it is a breaking change.
For now you should duplicate and deprecate imho.

WareditorGit:
Couldn't deprecate the enum though.

Frotty:
since you have to use the new one with the new function it should suffice.

PR #184, opened 2019-02-12
configurable for printing

PhoenixZeng:
1 maybe someone want a custom log-impl.
2 some short api for printing


Frotty:
`player.print` is missing hotdoc and you should probably use clean branches for PRs, but otherwise looks good 👍

Frotty:
Nice 🍬

PR #189, opened 2019-02-19
reworked PreloadIO to use 'BlzSetAbilityTooltip' instead of 'SetPlayerName'

Krypt0n:
This could be a better solution. Cause SetPlayerName is limit by MaxPlayerSlots and as for I my us I noteced less desyncs after around 100 tries in compare to the SetPlayerName but that you not effect the frame count. But basically it allows to save more data in smaller amount of files or even in only one File.

### DIFF
--- a/wurst/file/PreloadIO.wurst
+++ b/wurst/file/PreloadIO.wurst
@@ -12,13 +12,16 @@
 	you will need to prepend "Logs\\" to your paths.
 **/
 package PreloadIO
+
 import ErrorHandling
 import StringUtils
+import AbilityObjEditing
+import ObjectIds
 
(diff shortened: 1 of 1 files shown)


Frotty:
why close?

Krypt0n:
I am cheking a fork of samuel, and he already had this done there, I am merging those now and do several perfomance tests

Krypt0n:
I am using now samuels fork and did a few very small changes to adjuste the packet length if you get more then 99 preloads and for creating missing folders,
They work very fine for me even with larger amount of data, it is a bit slower and you get some delay, but there for no desyncs due to sync call frames limitation due to 1.30.2 patch
https://github.com/Krypt0n/WurstStdlib2

Issue #190, opened 2019-02-19
adding small letter to a linkedList says it has capital letter as well


Krypt0n:
![unbenannt](https://user-images.githubusercontent.com/640169/53050681-375d9a80-349a-11e9-9817-822e0ed2d6c7.PNG)

peq:
Moving to StdLib, because stringToIndex is implemented here (TypeCasting.wurst). The current version is just an approximation using StringHash.

Frotty:
and `StringHash` ignores casing.
Solution: new generics?

peq:
Just use StringHash like a normal hash function that might have collisions. So the toindex function should do something like linear probing.

Krypt0n:
I can work on a new StringHash for indexing if that is a point.

Frotty:
> Just use StringHash like a normal hash function that might have collisions. So the toindex function should do something like linear probing

I don't rly see the point of this breaking change, if new generics will just save the string as is without any type casting. Or am I missing something?

peq:
> I don't rly see the point of this breaking change,

It's not a (big) breaking change since it only behaves differently if there are collisions and noone should depend on what number is returned by `stringToIndex`. I hope nobody abuses this bug to have case insensitive string comparison.

>  if new generics will just save the string as is without any type casting. Or am I missing something?

Yes, but you can still use the old generics and they should work. 

For LinkedList, we could just change `public class LinkedList<T>` to `public class LinkedList<T:>` and it will work without changing much, but for things like `HashMap` the new generics are not yet expressive enough.

Krypt0n:
Yes I agree, it is just a thing I noticed, and thought it could be mentioned.

PR #191, opened 2019-02-20
Add linear probing to stringToIndex

peq:
See #190 

StringHash is not unique, so use linear probing to resolve colliding hashes.


Frotty:
Nice 🍰

PR #193, opened 2019-02-23
Move unit.isAliveTrick to widget.isAliveTrick

frederikaalund:
The trick applies to widgets in general and not just units.

Note that there is a technical difference in the implementation:
`unit.isAliveTrick` used the `GetUnitState` native whereas `widget.isAliveTrick`
uses the `GetWidgetLife` native. I'm unaware if there are any differences
in the output of said native functions. I assume not.

### DIFF
--- a/wurst/_handles/Unit.wurst
+++ b/wurst/_handles/Unit.wurst
@@ -259,10 +259,6 @@ public function unit.isInventoryFull() returns boolean
 public function unit.isAlive() returns boolean
 	return UnitAlive(this)
 
-/** Checks if the unit is alive by testing current HP > .405 */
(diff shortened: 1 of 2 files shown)


Frotty:
As I see it this change is already in https://github.com/wurstscript/WurstStdlib2/pull/192/files
Perhaps revert the commit there.

frederikaalund:
> Perhaps revert the commit there.

Done. I must have forgotten to create a branch for the PR first.

Frotty:
> Done. I must have forgotten to create a branch for the PR first.

np. You didn't have to open a new PR.
Anyway nice catch with this, thanks 🍻

PR #194, opened 2019-02-23
Generalize unit.damageTarget to target widgets

frederikaalund:
The previous behaviour was to only accept unit targets.
This change is backward-compatible.

The underlying native

    native UnitDamageTarget             takes unit whichUnit, widget target, real amount, boolean attack, boolean ranged, attacktype attackType, damagetype damageType, weapontype weaponType returns boolean

accepts `widget` targets so I can't think of a reason that the wrapper function `unit.damageTarget` shouldn't do so as well.

This pull request is an updated version of https://github.com/wurstscript/WurstStdlib2/pull/192 without the unrelated `isAliveTrick` changes.


Frotty:
Yea makes perfect sense, thanks a bunch 🍭

PR #198, opened 2019-02-25
Fix AbilityDefinitionPoisonAttack

Rhazarian:
"Poi4" field is actually an integer field, if you use a string it's simply ignored and acts the same way if the value was 0, even if the string could be parsed as an integer


Frotty:
Nice. Would you mind adding an enum to commons or so that wraps stacking types?

Rhazarian:
@Frotty ?

Frotty:
Yea, nice 👍

PR #205, opened 2019-03-11
Add some more extension functions for units

Nichts:


Frotty:
is there any benefit to `IsUnitInRangeXY` ?
Or just to have it wrapped?

Otherwise nice additions 👍

Nichts:
> is there any benefit to `IsUnitInRangeXY` ?
> Or just to have it wrapped?
> 
> Otherwise nice additions 👍

Just to have it wrapped - makes the code look more clean and consistent

Frotty:
> Just to have it wrapped - makes the code look more clean and consistent

Well my point is that you probably shouldn't ever use this native, but `.getPos().distanceToSq(..)` instead, since it could theoretically be faster if the native uses non squared distance and isn't intrinsically faster. It would also be more consistent and better style imo.

Nichts:
It should generally be faster or as fast (function calls in jass are expensive so getting the position in an extra call is probably not worth it).

But more importantly isunitinrange natives also consider the units collision size which yields slightly different results and is required if you try to emulate spells

Frotty:
> function calls in jass are expensive

Do we have any hard data on that? Imo most ppl are heavily exaggerating the performance implications of Jass. Also, whenever I had performance problems, there were a plethora of options I could optimize, before going into low level stuff like saving a few local variables or function calls.
I might be wrong in some edge cases, but I don't condone such 'proforma' decisions resulting in bad code style.  

> But more importantly isunitinrange natives also consider the units collision size which yields slightly different results and is required if you try to emulate spells

I think that is actually one of the only real differences/benefits, and thus should be noted in the hot doc.
It is def. not required, as you can do the collision check yourself, your only need to know the maximum collision size any unit in your map has. This has been done since RoC.

Frotty:
Nice, thanks 🌮

Nichts:
> > function calls in jass are expensive
> 
> Do we have any hard data on that? Imo most ppl are heavily exaggerating the performance implications of Jass. Also, whenever I had performance problems, there were a plethora of options I could optimize, before going into low level stuff like saving a few local variables or function calls.
> I might be wrong in some edge cases, but I don't condone such 'proforma' decisions resulting in bad code style.

I brought it up because you mentioned performance. This is one of the rare cases I actually think this is relevant, because functions like this are usually used inside loops along with EnumUnitsInRange potentially looping over massive amount of units.

> > But more importantly isunitinrange natives also consider the units collision size which yields slightly different results and is required if you try to emulate spells
> 
> I think that is actually one of the only real differences/benefits, and thus should be noted in the hot doc.
> It is def. not required, as you can do the collision check yourself, you only need to know the maximum collision size any unit in your map has. This has been done for a very long time.

I always do EnumUnitsInRange(AOE + MAX_COLLISION_SIZE) & loop with IsUnitInRangeXY(unit, x, y, AOE), because any other method I know either either hits small units more often or large units less often than it should. All information on this is actually quite old so maybe there is a better option I haven't found yet though 😕

PR #212, opened 2019-03-21
DamageEvent

WareditorGit:
A light-weight damage event package that use the new damage event native.
It's missing the 'detect native spells' part.


Frotty:
Thanks for committing, but this needs more info which should also be included as hot doc at top.
Ideally also a page for the [stdlib doc](https://wurstlang.org/stdlib.html) 
What exactly means "A light-weight damage event package" and does this replace DamageType?
It also does modification like https://github.com/wurstscript/WurstStdlib2/pull/60 ?

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,238 @@
...
+        return getDamageTypeName(t) + " Damage"
+    else if e != null 
+        return e.getName() + " Damage"
+    return "Damage"
+
+
+/* INIT */
+
+public function initializeDamageEventSystem()
+    onUnitIndex( () -> EventListener.add(getIndexingUnit(), EVENT_UNIT_DAMAGED, () -> DamageEvent.onDamage()))
```
Frotty: Any particular reason for this?
Why not just use `EventListener.add(EVENT_UNIT_DAMAGED) ->`?

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,238 @@
...
+
+
+/* CONFIGURATION */
+
+@configurable constant DETECT_NATIVE_ABILITIES = false // Not implemented
+
+@configurable public constant DAMAGE_ELEMENT_PHYSICAL = new DamageElement("Physical")
+
+
+/* DAMAGE TYPE */
```
Frotty: Please add actual hotdoc `/** */` to these top level declarations, explaining what they do / are used for.
Right now these comments seem a bit superfluous.

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,238 @@
...
+                nextDamageType = PHYSICAL
+                nextDamageElement = DAMAGE_ELEMENT_PHYSICAL 
+
+        let dmg = new DamageInstance(nextDamageId, GetEventDamageSource(), GetTriggerUnit(), GetEventDamage(), nextDamageType, nextDamageElement)
+        
+        nextDamageId = 0 
+        nextDamageType = UNKNOWN
+        nextDamageElement = null
+        
+        for i=0 to maxPriority
```
Frotty: i = 0

WareditorGit:
> It also does modification like #60 ?
Can you please tell me what exact modifications you have in mind? 

>what about circular damage events 
Imo it's a design issue. The system provides everything you need to design damage pipelines that evade infinite cycles. Providing protection is too much overhead.

>listeners proccing multiple times for the same damage
Haven't encountered this issue when using this system.

>I suppose you won't be implementing that part ? 
I suppose I will have to since it's can be useful.

Frotty:
> Can you please tell me what exact modifications you have in mind?

Changing amount (should be possible via set, maybe add another comfort wrapper) and preventing a unit from dying if it receives a possibly lethal hit when on low HP (back then this used to be done with a health gain ability, but with new native it's not need?).

> Imo it's a design issue. The system provides everything you need to design damage pipelines that evade infinite cycles. Providing protection is too much overhead.

I agree.

> Haven't encountered this issue when using this system.

Iirc it happened when changing the dealt damage through other modifiers - but this should be solved by priority. 

> I suppose I will have to since it's can be useful.

For backwards compatibility reasons it would be nice if DamageType would remain as a package for now. Perhaps you could `import public DamageEvent` into `DamageType` and replace everything except the exposed API (e.g. `setNextDamageCode()` in `dealCodeDamage`).
What do you think?

WareditorGit:
>preventing a unit from dying if it receives a possibly lethal hit when on low HP

Yeah it's not needed anymore.

> For backwards compatibility reasons it would be nice if DamageType would remain as a package for now. 

I will try and see what I can do but there is some conflicting data like the DamageType enum.

Frotty:
> I will try and see what I can do but there is some conflicting data like the DamageType enum.

Iirc both enums have 4 entries, so you could rename yours and remove the one from DamageType. Meaning seems pretty similar?

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,313 @@
+/** 
+    This package provides a light-weight damage detection system with priority-ordered listeners. 
+
+    **IMPORTANT** 
+        The system **must** be initialized using: initializeDamageEventSystem()
```
Frotty: does this have a particular reason?
What speaks against just doing it at init? (init blocks in external code is only executed when the package is imported directly or transitively in one of your packages)

WareditorGit: I had some issues with it not initializing when using classic init so I made the init manual at the time.  I will change it back to normal init and fill an issue if it happens again.

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,437 @@
...
+    CODE    // From user code
+    NULLED  // For backward compatibility -> has no utility
+    UNKNOWN         
+
+
+/* DAMAGE ELEMENT */
+
+/** Each damage instance can have a DamageElement associated to it. */
+public class DamageElement 
+    // Would be nice if we could @configurable a class
```
Frotty: ?!

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,437 @@
...
+    UNKNOWN         
+
+
+/* DAMAGE ELEMENT */
+
+/** Each damage instance can have a DamageElement associated to it. */
+public class DamageElement 
+    // Would be nice if we could @configurable a class 
+    protected string name 
+    protected colorA color = colorA(255, 255, 255, 255)
```
Frotty: use `COLOR_WHITE`.

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,437 @@
...
+import AbilityObjEditing
+import ClosureEvents
+import ObjectIdGenerator
+import OnUnitEnterLeave
+
+/* CONFIGURATION */
+
+/** If true, native abilities that deal damage are passed with DamageType SPELL.
+    Also keep in mind if true:
+        > Spell damage is converted into negative damage values without this system, so do not use any other damage detection method.
```
Frotty: repeated from above, replace with short remark.

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,437 @@
...
+    ATTACK  // From auto attacks or native abilities if DETECT_NATIVE_ABILITIES is false (PHYSICAL)
+    SPELL   // From native abilities (NATIVE_ABILITY)
+    CODE    // From user code
+    NULLED  // For backward compatibility -> has no utility
+    UNKNOWN         
+
+
+/* DAMAGE ELEMENT */
+
+/** Each damage instance can have a DamageElement associated to it. */
```
Frotty: Okay, but that doesn't tell me what the DamageElement does

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,437 @@
...
+                    print("This damage is from a native ability.")
+                case CODE 
+                    print("This damage is from code.")
+                default
+
+    Each damage instance can have an id and/or an element:
+
+        setNextEventDamageFromCode()
+        setNextEventDamageId('A000')
+        setNextEventDamageElement(FIRE_ELEMENT)
```
Frotty: I don't really get the point of the element. Is it like a custom ' type' ? What's the benefit compared to using `DAMAGE_TYPE_FIRE` e.g.? For texttags?

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,437 @@
...
+            print("This fires before.")
+
+    Each damage instance has a DamageType associated to it.
+    CODE **must** be declared by the user using DamageEvent.setNextDamageFromCode() before dealing damage.
+    ATTACK and SPELL are detected by the system (if DETECT_NATIVE_ABILITIES is false then native abilities will be detected as ATTACK)
+
+        addDamageListener() -> 
+            switch getEventDamageType() 
+                case ATTACK 
+                    print("This damage is from an auto-attack.")
```
Frotty: "auto-attack" kinda implies it wouldn't happen on user ordered attack

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,437 @@
...
+                case SPELL
+                    print("This damage is from a native ability.")
+                case CODE 
+                    print("This damage is from code.")
+                default
+
+    Each damage instance can have an id and/or an element:
```
Frotty: Cool, for what purpose? (i can figure, but explain for new devs)

wurst/event/DamageEvent.wurst
WareditorGit: You can't (for now) know the damageType of a damage instance so having an element is the way to create categories of damage instances for your map. Even when we will be able to access the damageType, having the power to have custom elements seems useful to me.

wurst/event/DamageEvent.wurst
WareditorGit: What would be the better term then? "unit's attack"?

wurst/event/DamageEvent.wurst
WareditorGit: DamageElement is a way to create categories for your damage instances. The settings of those categories are tied to the map's gameplay so being able to configure the class in some ways would be nice. For now, people can just extends from it.

WareditorGit:
>maybe add another comfort wrapper

>I'm not too sure if all those wrappers are really necessary thinking .

Yeah I kinda said fuck it and wrapped the whole thing...

wurst/event/DamageEvent.wurst
Frotty: Yea okay. Comment shouldn't be part of PR though.
Regarding `DamageElement` I think it's a good thing, the naming and usage just wasn't intuitive to me.
Element also to me implies that an attack can maybe have multiple elements.
Perhaps `CustomDamageType` or similar would be better. Element could be fine with updated doc.
Regarding doc, I think I would slim it down a bit and move examples and extra stuff into stdlib doc page: https://wurstlang.org/stdlib.html

Looking good 👍 
I think you can get rid of the class and just use a tuple for it though?

Frotty:
Btw, with "Wrappers" I didn't mean wrapping the static functions of DamageEvent, that seems superfluous.
But adding e.g. "addAmount" and "subAmount" wrappers for "setAmount" so the user don't has to do that part.

WareditorGit:
Ah that makes sense...
So should I remove them?

Frotty:
> So should I remove them?

Yeah, I think they would only make sense if they have some actual function in them.

If you fix that and last comment, I think it is mergable 🍰 
I will test it in my maps before though.

Frotty:
the deprecated annotations still contain old func names `@deprecated("Use GetEventDamageType() instead.")`.

Frotty:
Cool, LGTM. And for everything except configurable var it seems backwards compatible.
Thanks for your patience 👍

wurst/event/DamageEvent.wurst
```diff
@@ -0,0 +1,385 @@
...
+    /* LISTENERS */
+    /** Add a damage event listener.
+        If The order of firing is important, use addListener(priority, listener) */
+    static function addListener(DamageListener listener) returns DamageListener
+        return addListener(maxPriority, listener)
+
+    /** Add a damage event listener with a given priority.
+        Listeners of different priorities fire from the lowest priority to the highest priority added. 
+        Listeners of the same priority fire by order of addition (FIFO) */
+    static function addListener(int priority, DamageListener listener) returns DamageListener
```
Cokemonkey11: @WareditorGit 

Thanks for this. Your library is now the only one that properly handles 0-damage-events in 1.31.x (and I don't plan to fix DamageType or DamageDetection)

These docs need to describe constraints for priority. What is the max value, and what is the minimum one? The first thing I tried to do when using the priority management is set `-1`, which obviously won't work.

Issue #187, opened 2019-02-16
allow non standard paths for icon of dummy buffs


Krypt0n:
https://github.com/wurstscript/WurstStdlib2/blob/679af4db1001b1d6f7f4fd30731c96977581c354/wurst/objediting/BuffObjEditing.wurst#L30

right now if you want to use that function the path icon path must start with
"ReplaceableTextures" and well at least for me I don't  want to use that folder for everything

Frotty:
Then you want green icons? 🤔

Krypt0n:
well tbh that's a bit shitty then you have 
ReplaceableTextures\\CommandButtonsDisabled
for those
but having all icon for a project in 2 folders is a bit messy
a bit sad that only ReplaceableTextures\\CommandButtons can be subfoldered

Frotty:
Iirc you have to use those folders to avoid green icons, or did that change? 
I suppose it shouldn't be mandatory, what is your suggestion to avoid a breaking change?

Krypt0n:
well you only need that folder for the DIS, cause wc3 will just get all disable buttons (top avoid green icon thus) from "ReplaceableTextures\CommandButtonsDisabled\"
Not sure if it makes sense to have subfolder structure for "normal" icons but not for the "disabled" icons

I can ask some dev to change that in the future, so DIS- button cann go in any folder ...

Frotty:
Ah okay, so you could have an `imports/customfolder/BTNicon.blp` icon and wc3 will try to resolve the disabled variant from `imports/ReplaceableTextures/CommandButtonsDisabled/DISBTNicon.blp` ?
But for disabled only the root `CommandButtonsDisabled` is checked?

Krypt0n:
Yeah correctly

Frotty:
Okay so the only elegant non breaking way I can think about is prefixing your paths with some char like `@` which the preset function will detect and not add the `ReplaceableTextures` prefix, just remove the `@`.
Do you have another idea?

PR #213, opened 2019-03-22
added support for custompath for bufficons

Krypt0n:
for https://github.com/wurstscript/WurstStdlib2/issues/187

### DIFF
--- a/wurst/objediting/BuffObjEditing.wurst
+++ b/wurst/objediting/BuffObjEditing.wurst
@@ -74,6 +74,10 @@ public class BuffDefinition
 	function setIconNormal(int level, string value)
 		def.setLvlDataString("fart",  level, 0,  value)
 
+	/** 
+	 * Set the icon's path for the buff obj
+     * @param string name - name of the icon file, add "@!!" as prefix to overide default path by custom path
+	*/
 	function setIcon(string name)
 		string s = name
 		if (name.length() < 20 or not name.startsWith("ReplaceableTextures")) and not name.startsWith("@!!")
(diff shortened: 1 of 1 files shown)


Frotty:
That doesn't rly change the buff issue?

Krypt0n:
why not?

Frotty:
`createDummyBuffObject` uses `setIcon` and you cannot access the buff object afterwards?

Krypt0n:
ah yeah, its cause i have custom function for createDummyBuffObject ,  not sure if you want to have another function for createDummyBuffObject  with overloaded paramters

Frotty:
Which is why I suggested the solution I did.

Krypt0n:
Yeah I know, just thought its not that clean. I mean its just a note, the whoel thing is not a big problem at all.

Frotty:
Sure.
`function setIcon(string name)` should just reference the new one then tough.

Krypt0n:
Not sure if you want to have it in a configurable member, or just as it is now, I do not really care^^

Frotty:
I think it's fine, but should be added to hotdoc and perhaps other places? 🤧

Frotty:
will you add this feature to the function's hotdoc?

Krypt0n:
yeah sure, sry was away a few days

wurst/objediting/BuffObjEditing.wurst
```diff
@@ -74,9 +74,13 @@ public class BuffDefinition
 	function setIconNormal(int level, string value)
 		def.setLvlDataString("fart",  level, 0,  value)
 
+	/** 
+	 * Set the icon's path for the buff obj
+     * @param string name - name of the icon file, add "@!!" as prefix to overide default path by custom path
```
Frotty: `@param string name`
This is no part of the hotdoc specification. Please remove and use normal description.

Frotty:
Alright, this can't hurt imho. 👍

PR #218, opened 2019-03-30
More accurate forUnitsInRange()

WareditorGit:
Modified forUnitsInRange() to be more accurate - taking the units' collisions into consideration.

### DIFF
--- a/wurst/closures/ClosureForGroups.wurst
+++ b/wurst/closures/ClosureForGroups.wurst
@@ -94,9 +94,9 @@ public function forUnitsInRectCounted(rect r, int count, ForGroupCallback c)
 /** Executes the given closure for every unit in range of the given position */
 public function forUnitsInRange(vec2 pos, real radius, ForGroupCallback c)
 	pushCallback(c)
-	GroupEnumUnitsInRange(ENUM_GROUP, pos.x, pos.y, radius+MAX_COLLISION_SIZE, null)
+	GroupEnumUnitsInRange(ENUM_GROUP, pos.x, pos.y, radius + MAX_COLLISION_SIZE, null)
 	for u from ENUM_GROUP
-		if IsUnitInRangeXY(u, pos.x, pos.y, radius) //u.isInRangeOf(pos, radius)
+		if IsUnitInRangeXY(u, pos.x, pos.y, radius)
 			c.callback(u)
 	popCallback()
 
--- a/wurst/closures/ClosureForGroups.wurst
+++ b/wurst/closures/ClosureForGroups.wurst
@@ -94,7 +94,10 @@ public function forUnitsInRectCounted(rect r, int count, ForGroupCallback c)
 /** Executes the given closure for every unit in range of the given position */
 public function forUnitsInRange(vec2 pos, real radius, ForGroupCallback c)
 	pushCallback(c)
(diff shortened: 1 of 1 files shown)


Frotty:
Hi,
spacing is missing around binary operator, and what about the comment?
I didn't implement this initially iirc to leave the handling of collision size up to user and because the MAX_SIZE var thing isn't great.

WareditorGit:
Forgot to remove the comment (I used to use a native I added to unit - might do a PR for them later).

I agree it isn't the prettiest solution but I think it will pick units that the user would except to be picked. I assume some users don't even know about this trick and will wonder why the forUnitsInRange() sometimes doesn't pick some units that should be picked.

Frotty:
> Forgot to remove the comment (I used to use a native I added to unit - might do a PR for them later).

`unit.isInRange(vec2 pos, real distance) returns bool` already exists.

> I agree it isn't the prettiest solution but I think it will pick units that the user would except to be picked. I assume some users don't even know about this 'trick' and will wonder why the forUnitsInRange() sometimes doesn't pick some units that should be picked.

Yea I suppose. Inb4 worse performance though :P

WareditorGit:
> already exists.

Right but iirc at the time it didn't exist.

> worse performance though

If someone thinks it might be a big issue, can always add a config bool to toogle it on and off.

Frotty:
will u change to using ext funcs?

wurst/closures/ClosureForGroups.wurst
```diff
@@ -94,7 +94,10 @@ public function forUnitsInRectCounted(rect r, int count, ForGroupCallback c)
 /** Executes the given closure for every unit in range of the given position */
 public function forUnitsInRange(vec2 pos, real radius, ForGroupCallback c)
 	pushCallback(c)
-	GroupEnumUnitsInRange(DUMMY_GROUP, pos.x, pos.y, radius, filter)
+	GroupEnumUnitsInRange(ENUM_GROUP, pos.x, pos.y, radius + MAX_COLLISION_SIZE, null)
+	for u from ENUM_GROUP
```
Nichts: Why not have a filter for this (should be faster)

wurst/closures/ClosureForGroups.wurst
```diff
@@ -94,7 +94,10 @@ public function forUnitsInRectCounted(rect r, int count, ForGroupCallback c)
 /** Executes the given closure for every unit in range of the given position */
 public function forUnitsInRange(vec2 pos, real radius, ForGroupCallback c)
```
Nichts: This should probably be a new function with a different name to not break existing code (I currently rely on the fact that this does not do the collision size correction automatically and call it with radius + MAX_COLLISION_SIZE - this would effectively cause a lot more units to be grouped and then would be checked twice - once by this code once by mine)

Frotty:
Okay, so we won't change this std func for backwards compat I suppose.
What about a configurable constant to decide whether to apply the extra range stuff or not?

wurst/closures/ClosureForGroups.wurst
WareditorGit: Because that would be extra global variables to pass to the filter and I am not a fan of that. Not sure if it's faster too.

WareditorGit:
Making a new pr for this

PR #219, opened 2019-04-04
Define standard annotations

peq:
Since Wurst will now gives a warning for undefined annotations, this PR adds the predefined annotations to the standard library in `Annotations.wurst`.


Frotty:
👍

PR #220, opened 2019-04-11
Removed useless lines.

WareditorGit:
Removed useless lines in removeListener().


Frotty:
how did we miss that 😅 🍭

PR #221, opened 2019-04-15
Added preview links to the icons as reference.

apsyll:
Using the website"https://wowwiki.fandom.com/wiki/WC3_Icons/Icon_List" to have preview images for the icons.
you could use the mentioned extension in the editor to see the icons inside it.


apsyll:
I also noticed some icons are double and the second time it the variable has a 1 added other than that is nothing changed tho:
static constant dISBTNShadowPact = "ReplaceableTextures\\CommandButtonsDisabled\\DISBTNShadowPact.blp"
	static constant dISBTNShadowPact1 = "ReplaceableTextures\\CommandButtonsDisabled\\DISBTNShadowPact.blp"

Frotty:
Thanks for the contribution!
I'm not really a big fan of this solution tbh 🤔. Can you add a screenshot of how this looks?
We wanted to add texture/model preview that loads from game archives for some time - either natively in vscode, or as code lens that opens some app. I would prefer one of those.

apsyll:
Here would be a screenshot with the extension.
screenshot

Frotty:
> you could use the mentioned extension in the editor to see the icons inside it.

which one? Without the extension it's not doing anything?
I suppose it doesn't show icons of referenced constants, so you have to ctrl+click into the package?

I don't think it's ideal, but might be worth merging before the "real" impl.

apsyll:
The extension is in the code, in the first comment.
Sorry i'm on my mobile for the next weeks.

PR #222, opened 2019-04-17
Projectile class

GetLocalPlayer:
One of the most common tasks I suppose.

### DIFF
--- a/wurst/util/Projectile/Projectile.wurst
+++ b/wurst/util/Projectile/Projectile.wurst
@@ -1,0 +1,219 @@
+package Projectile
+import NoWurst
+import LinkedListModule
+import Timer
+import Effect
+import Interpolation
+import Vectors
+import Quaternion
+import Maths
+import Colors
+
+constant EPSILON = 0.0001
+
+public class Projectile
+	use LinkedListModule
+	protected unit owner = null
+	protected effect model = null
+	protected var scale = 1.
+	protected var timeScale = 1.
+	protected var distance = 0.
+	protected var speed = 0.
+	protected var remain = 0.
+	protected var start = ZERO3
+	protected var control = ZERO3
+	protected var finish = ZERO3
+	protected var pos = ZERO3
+	protected var prevPos = ZERO3
+	protected var orientation = IDENTITYQ
+	protected var direction = vec3(1, 0, 0)
+	protected var steering = false
+	protected var paused = false
+	protected var belowTerrain =  true
+	protected var clr = COLOR_WHITE
+
+	private static let TIMER = CreateTimer()
+	private static let DELAY = 0.02
+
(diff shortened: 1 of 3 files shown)


Frotty:
Hi, thanks for the PR.
Unfortunately this is a somewhat big, unannounced PR. 🤔 
You should make an issue first if you plan something like this, so we can discuss the requirements and implementation details.

From a quick view it looks too bulky to me. The effect stuff should be handled outside - probably Fx replacement that uses effects, a tuple could work for that I think.
Stuff like steering should either be lightweight & extendable, or completely user space imo. The current impl works for 3d I suppose, but seems totally overkill if you're just using 2d missiles.

Also it should be possible to create projectiles without "finish" point.
Such a target vector should be passed to a function that makes use of it, e.g. `enableSteering`.

`TargetProjectile` seems like bad name choice if the normal one already has a "finish". And what you do is periodically setting that to a new point. Target should also most likely be `widget`.
I think this should either be part of the parent's steering or implemented as a module. Not really sure.
Again though, something like `forceMid` is way too specific. I've never seen this nor used something like this - either make it fully customization or leave to user space. Much more important imo for a homing projectile are different kinds of homing. This of course doesn't work if the static "updateSteering" in the parent.

`TimedProjectile` seems almost superfluous? Why is this not in the parent? Why is speed always 0?

Instead of `use LinkedListModule` and timer, use `TimedLoop` module.
---

In general you should add some package hotdoc with a small example snippet to elaborate on the use of the classes.
The naming of most identifiers is also not ideal.

> I'd feel like Wurst having constructor call-chain (like in Java or C#), but no :(

Just add a setup function, which is called from the constructors. Not rly any difference besides calling `setup(...)` instead of  `this(...)`. There is a [ticket](https://github.com/wurstscript/WurstScript/issues/413) though.

GetLocalPlayer:
Well, ok. Does than mean such package can be a part of stdlib?

Frotty:
sure

GetLocalPlayer:
> probably Fx replacement that uses effect
By the way, should `Fx` package be rewritten as a wrapper for `effect` type since such type of dummy is not required anymore in latest patch?

Frotty:
> By the way, should Fx package be rewritten as a wrapper for effect type since such type of dummy is not required anymore in latest patch?

Yeah. Of course we need to keep the existing package for backwards compatibility, and some things are apparently still broken for effects (?) But an effect replacement would be nifty.
Regarding the dummy model, it should probably be added to project generation as legacy option, so we can remove the model from the default project.

GetLocalPlayer:
Ok, it looks like something completely different from this. I close that PR until I come up with something to create an issue.

Nichts:
@GetLocalPlayer please do so I don't have to role my own :smile:

PR #224, opened 2019-04-23
New reflection natives

peq:
Added new pseudo-natives for debugging memory leaks.

see https://github.com/wurstscript/WurstScript/commit/4adc23008880d2f6305d9f7207722ae3ed97aa04


Frotty:
total allocation count is missing?

PR #226, opened 2019-04-25
Fix for TimedLoop module

WareditorGit:
Fixing a bug introduced by the last commit.


WareditorGit:
Why this failed :thinking: ?

Frotty:
pls update from master, travis ci sucks with docker apparently :/

Frotty:
> I always do that though ttay24
 I always checkout master from upstream, add my commit and pr

Yea, but I changed the CI settings on master,  so you need to update from master again, and pus hto this PR...

Frotty:
So rn timedloop is totally borked :/ ? I was feeling bad about that last update.
Thanks for catching it ❤️

PR #227, opened 2019-04-29
forUnitsInRange with collision filtering

WareditorGit:
Following #218 discussion, added a new forUnitsInRange with collision filtering.


Frotty:
Thanks, I missed this 👍

PR #236, opened 2019-05-09
Fixed a division by zero case.

WareditorGit:
Fixed a division by zero in the Unit package when a unit had zero max mana.


Frotty:
Nice catch. I only wonder whether it should be full mana since 0/0 is also 100% 🤔 
Maybe add to hot doc that it will be 0 in that case?

WareditorGit:
Yeah I thought about that too. Let me add a hotdoc.

Frotty:
nice 🍰

Issue #180, opened 2018-12-24
Assets library should have pathing maps (the building footprints)


Poruft:
Would be nice to be able to access building pathing maps via the Assets library. Not sure if it makes more sense to have:

`PathingMaps.guardtower` (for every building in the game)

or

`PathingMaps.twoByTwo` (for all the rectangular pathing maps)

The building-based option makes it easy for people who know which building they want to mimic, but it isn't as straightforward for people who just want a rectangle. Perhaps both would be neat.

Frotty:
@Poruft 🚁

PR #237, opened 2019-05-09
added static  class for attachmentpoints

Krypt0n:


Frotty:
Neat 👍 Now we just need #180 and a few others 😄

Krypt0n:
Oh, for SpecialAttachmentPoints the "," needs to be changed to " "

PR #238, opened 2019-05-09
Added new natives:

Donach:
SaveFrameHandle
LoadFrameHandle
BlzGetSpecialEffectScale
BlzSetSpecialEffectMatrixScale
BlzResetSpecialEffectMatrix
BlzSendSyncData
BlzEnableCursor
BlzSetMousePos
BlzGetMouseFocusUnit
BlzGetLocalClientWidth
BlzGetLocalClientHeight
BlzIsLocalClientActive
BlzEnableSelections
BlzIsSelectionEnabled
BlzIsSelectionCircleEnabled
BlzCameraSetupApplyForceDurationSmooth
BlzDisplayChatMessage
BlzGetLocale
BlzEnableTargetIndicator
BlzIsTargetIndicatorEnabled
BlzChangeMinimapTerrainTex


Donach:
Might change some names of the functions, I tried to match already existing naming, but some of the new functions have weird names...

wurst.build
```diff
@@ -1,4 +1,4 @@
 ---
```
Frotty: don't commit this

wurst/_handles/Camera.wurst
```diff
@@ -81,31 +81,34 @@ public function camerasetup.setDestPosition(vec2 pos, real duration)
 public function camerasetup.getDestPosition() returns vec2
 	return vec2(CameraSetupGetDestPositionX(this),CameraSetupGetDestPositionY(this))
 
-public function camerasetup.apply(boolean doPan, boolean panTimed)
```
Frotty: These boolean->bool changes don't belong in this PR.
Perhaps we should decide a project wide convention, but for now leave them.

wurst/_handles/Effect.wurst
```diff
@@ -206,6 +206,18 @@ public function effect.clearSubAnimations()
 public function effect.removeSubAnimation(subanimtype whichSubAnim)
 	BlzSpecialEffectRemoveSubAnimation(this, whichSubAnim)
 
+/** Get effect scale.*/
+public function effect.getScale() returns real
+	return BlzGetSpecialEffectScale(this)
+
+/** Set effect matrix scale*/
```
Frotty: why is it matrix but takes vector 😕

Frotty:
Nice, looking good 👍

wurst/_handles/Effect.wurst
Donach: Is there any three-int tuple? if not, I can create one then...

wurst/_handles/Camera.wurst
Donach: will revert it then

wurst/_handles/Effect.wurst
Krypt0n: maybe call it public function effect.setScale(vec3 scaleVector)
and forget about blizzard misnaming things
maybe it will be changed after ptr

Donach: Oh, just noticed, nvm my old comment, wil lchange the name a bit

Donach:
Overwritten my previous commit, should fix all things pointed out.

Donach:
This PR is dependent on this https://github.com/wurstscript/WurstScript/pull/848
When that gets pulled, the tests from this PR should start working properly.

Donach:
Since new functions are implemented in WurstScript, could you manually run the build task again?
Thanks!
@Frotty

Frotty:
It runs but your assertions fail:

Donach:
Updated, now it will still fail on assertTrue(g.getUnitAt(g.size()-1) == u2), but that is because there is wrong implementation on the getUnitAt function in Java, not my fault :)
This needs to be fixed first for this to pass the check, but even though, I'd say this can be merged.
@Frotty @Krypt0n - https://github.com/wurstscript/WurstScript/pull/848

Frotty:
Looking good, but I would change all the functions like
```
if a
  return b
return c
```
To ternaries like `a ? b : c`

Donach:
@Frotty Can this be merged?

Frotty:
Yes 😄 🍺

PR #241, opened 2019-05-10
.modify() method for LLIterator

Trokkin:
Prove me wrong in this, or merge.
Maybe there was a way that I didn't see of simultaneous iterating over several linked lists with the ability to modify all of them?

Flexboxgrid.wurst:335 from my other PR is an example of its use.


Frotty:
Yea, I told you in chat already that `#updateAll` exists, which is loop + modification.

Frotty:
So yeah, as I said, this is a valid addition, but you need to add it to the back iterator too.

Trokkin:
Eh, that's clearly going too far. I'm reverting it.

Side note, I don't understand why does it throw ambiguity errors - LLEntry is not a subtype of Linked List, neither vice versa.

Frotty:
🌞

PR #243, opened 2019-05-11
Added new extensions functions for handles

Donach:
Created UnitUtils
Added two new ClosureTimer types

Should not interfere with any existing map.


Frotty:
Objects mean wurst class instances, I suppose you mean handles?
I would separate stuff added to existing packages, and new packages, like "UnitUtils" into own PRs.
The smaller, the easier PRs are to review and the chance to merge mistakes is far lower.
Also for stuff that isn't intrinsically obviousy, add some explanation.
What is the point of `fourPointTuple`? Most of it seems like duplicates of the "isInPolygon" etc. functions already existing.

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,32 @@ public function destructable.setOccluderHeight(real height)
 public function destructable.getName() returns string
 	return GetDestructableName(this)
 
+// Extensions
+/** used to toggles a destructable gate open close state*/
+public function destructable.gateToggle()
+	if GetDestructableLife(this) > 0
+		ModifyGateBJ(bj_GATEOPERATION_CLOSE, this)
```
Trokkin: `ModifyGateBJ` is a BJ function and thus can be implemented directly in wurst.
If done so, it also would not need bj_GATEOPERATION_# "enum" to "switch" between different actions

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,32 @@ public function destructable.setOccluderHeight(real height)
...
+	ModifyGateBJ(bj_GATEOPERATION_OPEN, this)
+
+/** used to close a destructable gate*/
+public function destructable.gateClose()
+	ModifyGateBJ(bj_GATEOPERATION_DESTROY, this)
+	ModifyGateBJ(bj_GATEOPERATION_CLOSE, this)
+
+/** used to move a destructable elevator one step up*/
+public function destructable.elevatorUp()
+	ChangeElevatorHeight(this, GetElevatorHeight(this)+1)
```
Trokkin: Also BJ, and `ChangeElevatorHeight` has even more "switches". I believe it could be implemented much more efficient.

wurst/_handles/Effect.wurst
```diff
@@ -206,6 +208,30 @@ public function effect.clearSubAnimations()
 public function effect.removeSubAnimation(subanimtype whichSubAnim)
 	BlzSpecialEffectRemoveSubAnimation(this, whichSubAnim)
 
+
+// Extensions
+/** remove effect after "time" expires*/
+public function effect.destr(real time)
+	doAfter(time) ->
```
Trokkin: I don't expect this to be inside handle's own API. ClosureTimers is an advanced thing in wurst that adds quite an overhead due to how lambda and method overriding work. Such functions fit much more in separate packages, but then I think that they're too simple to add to stdlib, especially if user will want to do something else aside from destroying it.

Trokkin: Oh, this can actually reduce the amount of generated overhead because there's only one lambda for this sole purpose. Still, shouldn't be inside _handles/.

wurst/_handles/Item.wurst
```diff
@@ -122,3 +126,24 @@ public function item.getIconPath() returns string
 
 public function item.setIconPath(string path)
 	BlzSetItemIconPath(this, path)
+
+// Extensions
+/** Updates the item's name, tooltip, extendedtooltip, description or icon for the specific player.
+	If p == null, the item is updated for all players */
+public function item.updateItem(player p, string itemName, string itemTooltip, string itemExtendedTooltip, string itemDesc, string itemIcon) returns bool
```
Trokkin: IMHO, this shouldn't be local to player, and then maybe have a 'ForPlayer' counterpart.

wurst/_handles/Player.wurst
```diff
@@ -151,6 +151,55 @@ public function player.getTaxRate(player otherPlayer, playerstate whichResource)
 public function player.isRacePrefSet(racepreference preference) returns bool
 	return IsPlayerRacePrefSet(this, preference)
 
+/** Sets the maximum allowed hero amount for a player */
+public function player.setHeroMaxAllowed(int amount)
+	SetPlayerMaxHeroesAllowed(amount, this)
```
Trokkin: Also BJ.
But this one is oneliner, probably doesn't matter - if only wurst would automatically inline it.

wurst/_handles/Unit.wurst
```diff
@@ -283,6 +292,18 @@ public function unit.isInRange(vec2 pos, real distance) returns bool
 public function unit.isEnemyOf(player whichPlayer) returns bool
 	return IsUnitEnemy(this, whichPlayer)
 
+/** Checks whether two units are enemies */
+public function unit.isEnemyOf(unit u) returns bool
+	return this != null and u != null and IsUnitEnemy(this, u.getOwner())
```
Trokkin: You don't need to check `!= null` to get proper results. Warcraft handles such cases in his own correct way. And even if it does not, the check is not supposed to be here for consistency reasons.

wurst/_handles/primitives/Real.wurst
```diff
@@ -4,6 +4,8 @@ import NoWurst
 public constant REAL_MAX = 340282366920938000000000000000000000000.
 public constant REAL_MIN = -340282366920938000000000000000000000000.
 
+public tuple lineFormular(real m, real b)
```
Trokkin: Sounds like something very much belonging to vec2 package or somewhere near it.

wurst/closures/ClosureTimers.wurst
```diff
@@ -175,6 +202,73 @@ public abstract class CallbackCounted
 	ondestroy
 		t.release()
 
+/** Same as CallbackSingle but it does not get deleted on call - must be destroyed manually after usage! **/
+public abstract class CallbackSingleNoAutoDestroy
```
Trokkin: I doubt the concept of not self-destroying callbacks. Could you provide an example use case?

wurst/util/GroupUtils.wurst
```diff
@@ -97,6 +100,56 @@ function createGroups(int number)
 			+ "All newly created groups will be non-recyclable.")
 		shownMaxError = true
 
+/** used to get units in rect with filter ! caution group needs to be released using group.release() after usage !*/
```
Trokkin: GroupUtils is more about utilizing groups correctly than having utility functions for getting them, so these should be in another package, I guess.

wurst/util/GroupUtils.wurst
```diff
@@ -97,6 +100,56 @@ function createGroups(int number)
...
+public function rect.getUnitsInRect(boolexpr filter) returns group
+	let g = getGroup()
+	GroupEnumUnitsInRect(g, this, filter)
```
Trokkin: Why not `g.enumUnitsInRect(this, filter)`?

wurst/util/GroupUtils.wurst
Trokkin: Also, I think these functions are bad for standart, because they implicitly create groups, *even if you state this in hotdoc*. And IMHO they are easily replicated on the spot, providing much clearer code.

wurst/util/MapBounds.wurst
```diff
@@ -39,7 +39,6 @@ public function vec2.inBounds() returns boolean
 public function vec3.inBounds() returns boolean
 	return inBoundsXY(this.x, this.y)
 
-
```
Trokkin: what?
e: I mean, this edit isn't necessary at all, so this file will appear in the list of edited files for no reason.

wurst/util/UnitUtils.wurst
```diff
@@ -0,0 +1,120 @@
...
+public function unit.applyBuffTimed(buffTuple buffApplied, real duration)
+	this.addAbility(buffApplied.abilId)
+	this.setAbilityLevel(buffApplied.abilId, 1)
+	doAfter(duration) ->
+		if this != null
+			this.setAbilityLevel(buffApplied.abilId, 0)
+			this.removeAbility(buffApplied.abilId)
+			
+/** Increase Armor of a unit for a duration */
+public function unit.addArmorTimed(real armorAdd, real duration)
```
Trokkin: This and all other Timed things sounds like a very specific mechanic implementation.
e: aside from applyBuffTimed thing, that thing is good.

wurst/util/UnitUtils.wurst
```diff
@@ -0,0 +1,120 @@
...
+	SetUnitAnimation(this, anim)
+	if duration != -1
+		doAfter(duration) ->
+			ResetUnitAnimation(this)
+
+public class OnFinishUnitAnimation
+	unit u = null
+	construct(unit u)
+		this.u = u
+	function onFinish()
```
Trokkin: consider naming `finish`?

wurst/_handles/Item.wurst
Krypt0n: why not localplayer?

wurst/util/GroupUtils.wurst
Krypt0n: true, i already supposed to remove all fitler stuff from the PR

wurst/util/UnitUtils.wurst
Krypt0n: you say this is to specific but the "applyBuffTimed" is good it is the same more or less, in both are special mechanics, so either add both with PR or non i would say

wurst/_handles/Item.wurst
Trokkin: Guess to not include Player package, cause _handle folder is already pretty much interconnected, and including it here would create loop dependency.
But again, this does not need to be player specific.

Krypt0n: Yeah that is true ofc.
We used it for only localplayer changes thats why we had it this way, guess the localplayer should be removed, and localplayer condition be checked on usage
Have the same for Frameapi btw...

wurst/util/UnitUtils.wurst
Donach: As Krypton mentions, I'd see this easily connected with any ability.
Cast ability on ally, which applyBuffTimed(x) and also addArmorTimed(Y, x) on given unit. Ofc this is useful only for coded spells, but atleast player does not have to define such function himself.

wurst/closures/ClosureTimers.wurst
Donach: @Krypt0n might wanna answer the original intention? Is it still needed/usable?

wurst/_handles/Item.wurst
Donach: I don't really get the point.
The tooltip change can occur on both local player only, or for all players, and this functions covers both.
But I don't really mind having two separate functions for this.

wurst/_handles/Destructable.wurst
Donach: Tried to rewrite it, dunno if I got more efficient, most likely not (as the animation names are irregular, cannot really see how to make it faster.
Anyway rewritten this to Wurst.

wurst/_handles/primitives/Real.wurst
```diff
@@ -97,3 +97,5 @@ public function real.lerp(real target, real alpha) returns real
 /** Checks if this real is between low and high value */
 public function real.isBetween(real low, real high) returns bool
 	return this >= low and this <= high
+
```
Frotty: ?

wurst/util/MapBounds.wurst
Frotty: yeah should be fixed

Frotty:
Files changed: 11 makes this review slow.

wurst/closures/ClosureTimers.wurst
Krypt0n: Well we need it quite often to not destroy the callback, but might restart then later, once it should repeated all its calls.
Another point is for code cleanse, cause I personally was looking for cleanup stuff and used class objects after they are not needed at all, not after they got called and self cleaned, so code pattern - sorry C dev here haha.

Moreover if you have many objects and instantiated classes like few hundred thousands at the same time your are limited sadly to the jass max array size, so we use recycler and double resuse classes, while keeping there members initiated by only destroyed on recycle. Guess no need to put these classes to stdlib cause they are far beyond general use cases.

Frotty: > Well we need it quite often to not destroy the callback, but might restart then later, once it should repeated all its calls.

Then use a periodic callback which you can stop and start?!

wurst/_handles/primitives/Real.wurst
Donach: Missing new line at the end of file

Donach:
Rebase gone a bit wrong... Hope I did not make it more messy...
Anyway I pushed the requested changes and rmeoved the NoDestroy callbacks.
Should now comply to the comments to provided. Let me know if there is anything else to be changed.
Thanks @Frotty @Trokkin

wurst/_handles/Unit.wurst
```diff
@@ -178,6 +178,15 @@ public function unit.getMaxHP() returns real
 public function unit.getMaxMana() returns real
 	return this.getState(UNIT_STATE_MAX_MANA)
 
+/** Get units current missing helth. */
```
Frotty: helth -> health

wurst/_handles/Unit.wurst
```diff
@@ -178,6 +178,15 @@ public function unit.getMaxHP() returns real
...
+public function unit.getMissingHP() returns real
+	return this.getMaxHP()-this.getHP()
```
Frotty: add spacing around binary operators, see https://wurstlang.org/manual.html#coding-conventions

wurst/math/Vectors.wurst
```diff
@@ -20,6 +20,23 @@ public constant UP = vec2(0, 1)
 public constant LEFT = vec2(-1, 0)
 public constant DOWN = vec2(0, -1)
 
+// Line Operations
+public tuple lineFormular(real m, real b)
+
+/** return f(y) for a given point on a line)*/
+public function real.getXOnLine(lineFormular line) returns real
+	return (this-line.b)/line.m
```
Frotty: spacing

wurst/util/MapBounds.wurst
Frotty: Resolved? You remove an extra line here, yet in your packages, like UnitUtils, there are many more empty lines?

wurst/util/UnitUtils.wurst
```diff
@@ -0,0 +1,120 @@
+package UnitUtils
+import BuffObjEditing
+import ClosureTimers
+
+
+
+/*
+
+This file contains various utility functions for Unit, which require enhanced functionality like closures and cannot be defined in Unit.wurst
```
Frotty: Less whitespace here and format into two lines

wurst/util/UnitUtils.wurst
```diff
@@ -0,0 +1,120 @@
...
+
+*/
+
+@configurable public constant real FIELD_OF_VIEW_FRONT  = 115.00  // Real FoV is: 180 - 2x (FIELD_OF_VIEW-90)	Using a value of 115 generates a FoV of 130   (natural FoV is about 135 degrees)
+@configurable public constant real FIELD_OF_VIEW_BEHIND = 115.00
```
Frotty: why is this a constant and not a function parameter?

wurst/util/UnitUtils.wurst
```diff
@@ -0,0 +1,120 @@
...
+
+
+
+/*
+
+This file contains various utility functions for Unit, which require enhanced functionality like closures and cannot be defined in Unit.wurst
+
+*/
+
+@configurable public constant real FIELD_OF_VIEW_FRONT  = 115.00  // Real FoV is: 180 - 2x (FIELD_OF_VIEW-90)	Using a value of 115 generates a FoV of 130   (natural FoV is about 135 degrees)
```
Frotty: unused?

wurst/util/UnitUtils.wurst
```diff
@@ -0,0 +1,120 @@
...
+		this.addAbility(buffApplied.abilId)
+		this.setAbilityLevel(buffApplied.abilId, 1)
+
+		let targetUnit = this // saveback for closure
+		let maxCount   = (duration/interval).floor()+1	  //  last proc are is cut off 
+		doPeriodicallyCounted(interval, maxCount) cb ->	 // first proc will be applied after 1*interval
+			if targetUnit == null or not targetUnit.isAlive() or cb.isLast()
+				targetUnit.setAbilityLevel(buffApplied.abilId, 0)
+				targetUnit.removeAbility(buffApplied.abilId)
+				cb.stop()
```
Frotty: you don't need to stop before destroying I think

Frotty:
No worries, commits don't matter much as this will be squashed before merge. 👍 
One more thing: Please make the hotdoc formatting consistent, i.e. spacing before `/*` and `*/` and starting with a captial letter. Right now they are mixed.

wurst/_handles/Unit.wurst
Krypt0n: I always need to do this in his code :((( so annoying xD

wurst/util/UnitUtils.wurst
Krypt0n: well cause its mostly not needed to be changed, could be seen as some default value that makes sense in most common situations, but just in case someone wants to use different settings, could be parameter and the 115 added into the hotdoc i think

wurst/util/UnitUtils.wurst
Krypt0n: True was only needed in a old version of stdlib, where it did not work properly..

wurst/util/UnitUtils.wurst
Donach: Added isInFront

wurst/util/StringUtils.wurst
```diff
@@ -157,6 +126,101 @@ public function string.reduce(string reduced) returns string
...
+		s = s.substring(0, s.length()-1)
+	if paramCount > WC3LIMIT_FUNCTIONS_PARAMETER_COUNT
+		error(WC3LIMIT_FUNCTIONS_PARAMETER_COUNT_ERROR)
+	return s
+
+/** Checks whether a string is representing a digit or not.*/
+public function string.isDigit() returns bool
+	bool isDigit = false
+	for int i = 0 to this.length()-1
+		if this.charAt(i) == "0" or this.charAt(i) == "1" or this.charAt(i) == "2" or this.charAt(i) == "3" or this.charAt(i) == "4"
```
bosorioo: Saving the return of this.charAt in a variable would save a lot of useless calls here.  
Also, there is a simpler implementation to this consisting of calling S2I to convert the string to number and checking if it's different from 0. If the conversion returns 0, then you have to check if the string is actually "0", because any string that can't be converted to integer returns 0. Obvisouly this method requires some experimentation to see what happens/should happen when the string contains the minus operator or has spaces in it.

Donach: That is not entirely true.
If you have string like "241a14", the S2I function will actually convert this to 241. This function checks whether the full string can be considered a number, not just a part of it.
Your other suggestions are fine, I'll give it a look.

Donach:
There seems to be some problme with CircleCI, any idea what could be wrong?
@Frotty @Trokkin

Krypt0n:
@Donach  just resolve the conflicts ...
Capture

Donach:
Cannot do really more than this regadring the conflicts.
Can you just merge it manually @Frotty ?
It should be quite simple - just make sure to INCLUDE all the additions from my branch which is missing in the master.
Then this should be ready to go.

Thanks.

Frotty:
Why don't you just rebase?
Also still too many changes for one PR.

Krypt0n:
Yeah to much stuff here still, that should not be in the stdlib at all.

Donach:
Reverted couple of changes, cut the amount of code pushed, now contains mostly new functions + some nice extensions
Can you give it a look @Frotty ?

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,78 @@ public function destructable.setOccluderHeight(real height)
 public function destructable.getName() returns string
 	return GetDestructableName(this)
 
+// Extensions
+/** used to open a destructable gate*/
+public function destructable.gateOpen()
```
Cokemonkey11: I'm not sure this belongs here. Is it just a wrapper for `kill`?

It may be better to wrap `destructable` in a tuple type for gates `tuple gate(destructable killable)`

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,78 @@ public function destructable.setOccluderHeight(real height)
...
+	if (GetDestructableLife(this) > 0)
+		KillDestructable(this)
+
+/** used to close a destructable gate*/
+public function destructable.gateClose()
```
Cokemonkey11: Ditto the above - let's use tupletypes to add some type-safety for gates, elevators, switches.

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,78 @@ public function destructable.setOccluderHeight(real height)
...
+	this.gateOpen()
+	DestructableRestoreLife(this, GetDestructableMaxLife(this), true)
+	SetDestructableAnimation(this, "stand")
+
+/** used to close a destructable gate*/
+public function destructable.gateDestroy()
```
Cokemonkey11: ditto

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,78 @@ public function destructable.setOccluderHeight(real height)
...
+	this.gateOpen()
+	SetDestructableAnimation(this, "death")
+
+/** used to toggles a destructable gate open close state*/
+public function destructable.gateToggle()
```
Cokemonkey11: Ditto

I would prefer something like `setOpen(bool openness)`

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,78 @@ public function destructable.setOccluderHeight(real height)
...
+	if GetDestructableLife(this) > 0
+		this.gateClose()
+	else
+		this.gateOpen()
+
+constant let ELEVATOR_HEIGHT_MIN = 1
+constant let ELEVATOR_HEIGHT_MAX = 3		
+/** used to move a destructable elevator one step up*/
+public function destructable.elevatorUp()
```
Cokemonkey11: Let's declare `tuple elevator(destructable elevatable)`

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,78 @@ public function destructable.setOccluderHeight(real height)
...
+	this.setElevatorHeight(this.elevatorGetHeight()+1)
+
+/** used to move a destructable elevator one step down*/
+public function destructable.elevatorDown()
```
Cokemonkey11: ditto

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,78 @@ public function destructable.setOccluderHeight(real height)
...
+	this.setElevatorHeight(this.elevatorGetHeight()-1)
+
+/** used to move a destructable elevator one step down*/
```
Cokemonkey11: Comment false

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,78 @@ public function destructable.setOccluderHeight(real height)
...
+public function destructable.elevatorSetHeight(int height)
```
Cokemonkey11: ditto use tuple

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,78 @@ public function destructable.setOccluderHeight(real height)
...
+	this.setElevatorHeight(height)
+
+/** used to move a destructable elevator one step down*/
```
Cokemonkey11: Comment false

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,78 @@ public function destructable.setOccluderHeight(real height)
...
+public function destructable.elevatorGetHeight() returns int
+	let height = ELEVATOR_HEIGHT_MIN + (this.getOccluderHeight() / bj_CLIFFHEIGHT).toInt()
+	return height < ELEVATOR_HEIGHT_MIN or height > ELEVATOR_HEIGHT_MAX ? ELEVATOR_HEIGHT_MIN : height
```
Cokemonkey11: looks useful, don't throw this code away pls

wurst/_handles/Destructable.wurst
```diff
@@ -69,6 +69,78 @@ public function destructable.setOccluderHeight(real height)
...
+	return height < ELEVATOR_HEIGHT_MIN ? ELEVATOR_HEIGHT_MIN : (height > ELEVATOR_HEIGHT_MAX ? ELEVATOR_HEIGHT_MAX : height)
+
+function destructable.setElevatorHeight(int height)
+	let oldHeight = this.elevatorGetHeight()
+	let newHeight = getElevatorHeightBound(height)
+	this.setOccluderHeight(bj_CLIFFHEIGHT*(newHeight-ELEVATOR_HEIGHT_MIN))
+
+	var animationInstant = "stand"
+	var animationQueue = "stand"
+	switch (newHeight)
```
Cokemonkey11: very cool, but as above, needs tuple wrapper for type-safety

wurst/_handles/Item.wurst
```diff
@@ -15,6 +15,10 @@ public function item.remove()
 public function item.getTypeId() returns int
 	return GetItemTypeId(this)
 
+/** Returns "itemtype" of this item. Do not confuse it with itemTypeId. */
+public function item.getItemType() returns itemtype
```
Cokemonkey11: As `itemtype` is a type, can we call this `getItemtype()`? It will also aide in avoiding the concern you mention in the comment.

wurst/_handles/Item.wurst
```diff
@@ -161,3 +165,35 @@ public function item.setField(itemrealfield field, real value) returns bool
 
 public function item.setField(itemstringfield field, string value) returns bool	
 	return BlzSetItemStringField(this, field, value)
+
+// Extensions
```
Cokemonkey11: `// Extensions` ?

wurst/_handles/Item.wurst
```diff
@@ -161,3 +165,35 @@ public function item.setField(itemrealfield field, real value) returns bool
...
+/** Updates the item's name, tooltip, extendedtooltip, description or icon for the specific player.
+	If p == null, the item is updated for all players 
+	Returns false if no player is given. */
+public function item.updateItemForPlayer(player p, string itemName, string itemTooltip, string itemExtendedTooltip, string itemDesc, string itemIcon) returns bool
+	return p != null ? this.updateItem(p, itemName, itemTooltip, itemExtendedTooltip, itemDesc, itemIcon) : false
+
+/** Updates the item's name, tooltip, extendedtooltip, description or icon for the specific player.
+	If p == null, the item is updated for all players 
+	Returns false if no player is given. */
+public function item.updateItemForAll(string itemName, string itemTooltip, string itemExtendedTooltip, string itemDesc, string itemIcon) returns bool
```
Cokemonkey11: In other parts of the standad library we would just call this `item.updateDetails` and the other one `item.updateDetailsForPlayer`

wurst/_handles/Player.wurst
```diff
@@ -210,6 +210,55 @@ public function player.sendSyncData(string prefix, string data) returns bool
 public function player.showChatMessage(int recipient, string message)
 	BlzDisplayChatMessage(this, recipient, message)
 
+/** Sets the maximum allowed hero amount for a player */
+public function player.setHeroMaxAllowed(int amount)
+	SetPlayerTechMaxAllowed(this, 'HERO', amount)
+		
+/** Change tooltip (Name) of given ability for player locally.*/
+public function player.setAbilityTooltip(int abilityId, string abilityName, int abilityLevel)
```
Cokemonkey11: These ability-wide changes are probably more appropriate as methods of a new tuple type, `ability(int id)`

wurst/_handles/Unit.wurst
Cokemonkey11: third paramater could be `this.isEnemyOf(u.getOwner())`

wurst/_handles/Unit.wurst
```diff
@@ -283,6 +292,18 @@ public function unit.isInRange(vec2 pos, real distance) returns bool
 public function unit.isEnemyOf(player whichPlayer) returns bool
 	return IsUnitEnemy(this, whichPlayer)
 
+/** Checks whether two units are enemies */
+public function unit.isEnemyOf(unit u) returns bool
+	return this != null and u != null and IsUnitEnemy(this, u.getOwner())
+
+/** Checks whether a unit is an ally of the given player */
+public function unit.isAllyOf(player whichPlayer) returns bool
+	return this != null and IsUnitAlly(this, whichPlayer)
```
Cokemonkey11: ditto

wurst/_handles/Unit.wurst
```diff
@@ -283,6 +292,18 @@ public function unit.isInRange(vec2 pos, real distance) returns bool
...
+
+/** Checks whether two units are allies */
+public function unit.isAllyOf(unit u) returns bool
```
Cokemonkey11: ditto

wurst/util/EffectUtils.wurst
```diff
@@ -0,0 +1,27 @@
...
+
+/** Move effect to invisible part of map before removing it (which plays death animation).
+	Does not work on effects attached to units!
+*/
+public function effect.destrEx()
+	this..setColor(colorA(0, 0, 0, 0))
+		..setPos(vec3(boundMin.x, boundMin.y, 5000))
+		..setTime(1)
+		..setTimeScale(10)
+		..destr()
```
Cokemonkey11: Missing EoF LF

wurst/util/EffectUtils.wurst
```diff
@@ -0,0 +1,27 @@
+package EffectUtils
```
Cokemonkey11: Simple extension methods of `effect` should be in `handles/Effect.wurst`

wurst/util/EffectUtils.wurst
```diff
@@ -0,0 +1,27 @@
...
+// Extensions
+/** remove effect after "time" expires*/
+public function effect.destr(real time)
+	doAfter(time) ->	  
+		this.destr()
+
+/** After "time", move effect to invisible part of map before removing it (which plays death animation).
+	Does not work on effects attached to units!
+*/
+public function effect.destrEx(real time)
```
Cokemonkey11: Maybe call this `remove()`, since the point is to hide the death animation

wurst/util/UnitUtils.wurst
```diff
@@ -0,0 +1,60 @@
+package UnitUtils
+import BuffObjEditing
+import ClosureTimers
+
+
+
+/*	This file contains various utility functions for Unit, 
+	which require enhanced functionality like closures and cannot be defined in Unit.wurst */
```
Cokemonkey11: Is this something @Frotty  has suggested?

wurst/util/UnitUtils.wurst
```diff
@@ -0,0 +1,60 @@
...
+		if this != null
+			this.setAbilityLevel(buffApplied.abilId, 0)
+			this.removeAbility(buffApplied.abilId)
+			
+/** Increase Armor of a unit for a duration */
+public function unit.addArmorTimed(real armorAdd, real duration)
+	BlzSetUnitArmor(this, this.getArmor()+armorAdd)
+	doAfter(duration) ->
+		if this != null
+			BlzSetUnitArmor(this, this.getArmor()-armorAdd)
```
Cokemonkey11: This is not safe to multi-instanceability and would need a lock or guard

wurst/util/UnitUtils.wurst
```diff
@@ -0,0 +1,60 @@
...
+			
+/** Increase Damage of a unit for a duration */		
+public function unit.addDamageTimed(int damageAdd, real duration)
+	BlzSetUnitBaseDamage(this, this.getBaseDamage(1)+damageAdd, 1)
```
Cokemonkey11: As above, needs a lock or guard. This will break if a unit gets bonus damage calculated as percentage, and bonus damage calculated as a scalar, and then they're reversed FIFO.

wurst/util/UnitUtils.wurst
```diff
@@ -0,0 +1,60 @@
...
+	this.setAttackCooldown(newSpeed, weaponIndex)
+	return newSpeed
+   
+/** Change unit movement speed by given percentage. */
+public function unit.setMoveSpeedPercChange(real movespeedPercChange) returns real
+	let oldSpeed = this.getMoveSpeed()
+	let newSpeed = oldSpeed + (oldSpeed*movespeedPercChange)
+	this.setMoveSpeed(newSpeed)
+	return newSpeed
+
```
Cokemonkey11: wtf, is this a CR with no LF?

wurst/_handles/Destructable.wurst
Donach: Sarcasm?

wurst/_handles/Item.wurst
Donach: Can't find it in my workspace...

wurst/_handles/Destructable.wurst
Cokemonkey11: no not sarcasm

wurst/util/EffectUtils.wurst
Donach: Don't see any such in the Utils. One requires closures, the last requires MapBounds import, that should not be part of Handles I'd say...

wurst/util/UnitUtils.wurst
Donach: What about making these functions @configurable ?
Anyone could change these based on their needs. I'd say we hardly can cover all the possible cases inside Stdlib.

wurst/util/UnitUtils.wurst
Donach: CR? (Change Request -> Pull Request?)
LF?

wurst/util/UnitUtils.wurst
Cokemonkey11: I think this should be an external library if it's unsafe - or, at the very least - good documentation about limitations. I'll defer to Frotty.

wurst/util/UnitUtils.wurst
Cokemonkey11: CR = Carriage Return

LF = Line Feed

wurst/_handles/Player.wurst
Donach: I kinda cannot see any added benefit here.
Can you elaborate more?
Also there can't be tuple "ability" cause there is new native "ability"

wurst/util/UnitUtils.wurst
Donach: What about making these functions @configurable ?
Anyone could change these based on their needs. I'd say we hardly can cover all the possible cases inside Stdlib.

wurst/util/UnitUtils.wurst
Donach: I was told anything using closures should go into separate XUtils.wurst file, never to the handles file itself if that's what you're asking.

wurst/_handles/Player.wurst
Cokemonkey11: > I kinda cannot see any added benefit here.
> Can you elaborate more?

Type safety

namespace cleanliness (don't add stuff to `player`)

> Also there can't be tuple "ability" cause there is new native "ability"

Ok, maybe as extension-functions for `ability` then

wurst/util/UnitUtils.wurst
Cokemonkey11: Sounds good to me

wurst/util/UnitUtils.wurst
Cokemonkey11: I'm not a fan, but will defer to frotty. Prefer this in library

wurst/util/UnitUtils.wurst
```diff
@@ -0,0 +1,60 @@
...
+/** Returns true if this unit is behind target unit.
+    Uses configurable FIELD_OF_VIEW_BEHIND variable with default value of 115. */
+public function unit.isBehind(unit targetUnit) returns bool
+	let degreeDif = this.getFacingAngle().degrees() - targetUnit.getFacingAngle().degrees()
+	return -FIELD_OF_VIEW_BEHIND <= degreeDif and degreeDif <= FIELD_OF_VIEW_BEHIND
+
+/** Returns true if this unit is in front of target unit.
+	Uses configurable FIELD_OF_VIEW_FRONT variable with default value of 230. */
+public function unit.isInFront(unit targetUnit) returns bool
+	let degreeDif = this.getFacingAngle().degrees() - targetUnit.getFacingAngle().degrees()
```
Cokemonkey11: I don't think it makes sene to compare two units facing angle for `inFront`ness.

We really want to compare one unit's facing angle to another unit's position

Krypt0n: Yeah its totally bullshit, also is behind

wurst/util/UnitUtils.wurst
Krypt0n: Also unfinished  code here, that should not have been added to PR,
cause it should be added with statushandler same for any another change like armor, life etc.

wurst/util/UnitUtils.wurst
Donach: Removed these old functions

wurst/util/UnitUtils.wurst
Donach: removed

wurst/util/UnitUtils.wurst
Donach: removed

wurst/util/EffectUtils.wurst
Cokemonkey11: No problem I'm happy

wurst/util/UnitUtils.wurst
Cokemonkey11: Here's some code that will do it:

```
    public function angle.smallestDiff(angle other) returns angle
        return this.direction().dot(other.direction()).acos().asAngleRadians()

    public function angle.toVec2() returns vec2
        return vec2(0., 0.).polarOffset(this, 1.)

    public function angle.lt(angle other) returns bool
        return (this.radians() % (2. * PI)) < other.radians()

    public function angle.toString() returns string
        return (this.degrees() % 360.).toString(2)

    public function unit.isFacing(unit which, real maxRange, angle maxAng) returns bool
        let angBetween = this.getPos().angleTo(which.getPos())
        let distBetween = this.getPos().distanceToSq(which.getPos())
        return (angBetween.smallestDiff(this.getFacingAngle())).lt(maxAng) and distBetween < maxRange*maxRange
```

Donach:
Removed unsuitable functions.
Can this be approved @Cokemonkey11 @Frotty ?

wurst/util/UnitUtils.wurst
Krypt0n: Yeah that is still not what I was talking about, I have a bit more complex code, picking all units in front or behind or in specific direction of the unit / pos. While allowing some filter, and configurable for search pattern shape and sizes. Iam not 100% if that is a thing for lib as it is a bit more math

wurst/_handles/Unit.wurst
Cokemonkey11: > Warcraft handles such cases in his own correct way.

Ha! No it doesn't @Trokkin 

wc3 is awful on null checks for unit handles

But, maybe it should be removed, as map author's responsibility. I am not sure. @Frotty ?

wurst/util/UnitUtils.wurst
Cokemonkey11: That seems hardly different from this

wurst/util/EffectUtils.wurst
Frotty: The naming should be more consistent here.
`remove` should imply what it's doing, like `hiddenDestroy`.
The `destr` overload with a real param seems confusing, should be `destroyAfter` and `hiddenDestroyAfter` imo.

Frotty:
Yea looks cool and useful now!
Sorry for the delay 😅

Frotty:
Some neat additions, can finally merge 🎉 
Thanks a bunch 🍻

PR #244, opened 2019-05-11
Added new string extension functions

Donach:
Moved some functions from StringUtils to LinkedList and removed StingUtil import from ObjectIds to avoid cyclic dependency

### DIFF
--- a/wurst/data/LinkedList.wurst
+++ b/wurst/data/LinkedList.wurst
@@ -515,13 +515,7 @@ public function group.asList() returns LinkedList<unit>
 	this.forEachFrom() u ->
 		result.add(u)
 	return result
-
-public function LinkedList<T>.getLast<T>() returns T
-	T elem = null
-	if this.size() > 0
-		elem = this.get(this.size()-1)
-	return elem
-
+	
 /** Joins elements from a string list into one string using a separator */
 public function LinkedList<string>.joinBy(string separator) returns string
 	var joined = ""
--- a/wurst/data/LinkedList.wurst
+++ b/wurst/data/LinkedList.wurst
@@ -516,5 +516,46 @@ public function group.asList() returns LinkedList<unit>
 		result.add(u)
 	return result
 
+public function LinkedList<T>.getLast<T>() returns T
+	T elem = null
+	if this.size() > 0
+		elem = this.get(this.size()-1)
+	return elem
+
+/** Joins elements from a string list into one string using a separator */
+public function LinkedList<string>.joinBy(string separator) returns string
+	var joined = ""
+	var first = true
+	let iter = this.iterator()
+
+	for str from iter
+		if iter.hasNext() or not first
+			joined += str + separator
+		else
+			joined += str
+			first = false
+
+	iter.close()
+	return joined
+
+interface ToStringClosure<T>
+	function toString(T elem) returns string
+
+public function LinkedList<T>.joinBy<T>(ToStringClosure<T> cls, string separator) returns string
+	let iter = this.iterator()
+	LinkedList<string> strings = new LinkedList<string>
+
+	for elem from iter
+		strings.add(cls.toString(elem))
+
+	iter.close()
+	let result = strings.joinBy(separator)
+	destroy strings
+	return result
+
+/** Joins elements from a string list into one string */
+public function LinkedList<string>.join() returns string
+	return this.joinBy("")
+
 init
 	realToIndex(0.)
--- a/wurst/objediting/ObjectIds.wurst
+++ b/wurst/objediting/ObjectIds.wurst
@@ -4,7 +4,6 @@ import ErrorHandling
 import Wurstunit
 import Assets
 import LinkedList
-import StringUtils
 import TypeCasting
 import Annotations
 
--- a/wurst/util/StringUtils.wurst
+++ b/wurst/util/StringUtils.wurst
@@ -1,11 +1,12 @@
 package StringUtils
 import NoWurst
-import String
 import LinkedList
 import Texttag
 import Interpolation
 import TypeCasting
 import Annotations
+import ErrorHandling
+import ObjectIds
 
 /**
  * Credits go here: http://www.hiveworkshop.com/forums/jass-resources-412/snippet-ascii-190746/
@@ -17,6 +18,9 @@ int array width
 constant MAX_INDEX = '~' + 1
 constant DEFAULT_WIDTH = 0
 constant DEFAULT_STRING = ""
+constant int	 WC3LIMIT_FUNCTIONS_PARAMETER_COUNT  = 31 // Should be a global for whole WurstScript or StdLib at least
+constant int	 WC3LIMIT_OBJECTS_DESCRIPTION_LENGTH = 1023 // Should be a global for whole WurstScript or StdLib at least
+constant string  WC3LIMIT_FUNCTIONS_PARAMETER_COUNT_ERROR = "Maximum number of parameters reached!"
 
 public function texttag.center(vec3 pos, string message, real size) returns texttag
 	var str = message
@@ -101,41 +105,6 @@ public function string.split(string splitBy) returns LinkedList<string>
 		str = second
 	return list
 
-/** Joins elements from a string list into one string using a separator */
-public function LinkedList<string>.joinBy(string separator) returns string
-	var joined = ""
-	var first = true
-	let iter = this.iterator()
-
-	for str from iter
-		if iter.hasNext() or not first
-			joined += str + separator
-		else
-			joined += str
-			first = false
-
-	iter.close()
-	return joined
-
-interface ToStringClosure<T>
-	function toString(T elem) returns string
-
-public function LinkedList<T>.joinBy<T>(ToStringClosure<T> cls, string separator) returns string
-	let iter = this.iterator()
-	LinkedList<string> strings = new LinkedList<string>
-
-	for elem from iter
-		strings.add(cls.toString(elem))
-
-	iter.close()
-	let result = strings.joinBy(separator)
-	destroy strings
-	return result
-
-/** Joins elements from a string list into one string */
-public function LinkedList<string>.join() returns string
-	return this.joinBy("")
-
 /** Replaces consecutive repeats of the same string with a single instance
 	Turns "strstr" into "str" */
 public function string.reduce(string reduced) returns string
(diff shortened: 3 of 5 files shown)


wurst/data/LinkedList.wurst
```diff
@@ -516,5 +516,46 @@ public function group.asList() returns LinkedList<unit>
 		result.add(u)
 	return result
 
+public function LinkedList<T>.getLast<T>() returns T
```
Frotty: Why not use the LinkedList's method to get the last element which is O(1)?
The `.get` call is O(n)

wurst/objediting/ObjectIds.wurst
```diff
@@ -4,7 +4,6 @@ import ErrorHandling
 import Wurstunit
 import Assets
 import LinkedList
-import StringUtils
```
Frotty: You can't just do breaking things like this without any replacement.

wurst/util/StringUtils.wurst
```diff
@@ -17,6 +18,9 @@ int array width
 constant MAX_INDEX = '~' + 1
 constant DEFAULT_WIDTH = 0
 constant DEFAULT_STRING = ""
+constant int	 WC3LIMIT_FUNCTIONS_PARAMETER_COUNT  = 31 // Should be a global for whole WurstScript or StdLib at least
```
Frotty: No, this should be (and already is) inbuilt in the compiler.
You will get a warning if using more than 31 parameters.
Having this inside the vararg functions isn't a good idea imo. Especially if it's only certain ones.

wurst/util/StringUtils.wurst
```diff
@@ -17,6 +18,9 @@ int array width
...
+constant int	 WC3LIMIT_OBJECTS_DESCRIPTION_LENGTH = 1023 // Should be a global for whole WurstScript or StdLib at least
+constant string  WC3LIMIT_FUNCTIONS_PARAMETER_COUNT_ERROR = "Maximum number of parameters reached!"
```
Frotty: If you use constant, omit the type.

wurst/util/StringUtils.wurst
```diff
@@ -157,6 +126,101 @@ public function string.reduce(string reduced) returns string
...
+		s += id.toRawCode()
+		s += ","
+	if s != ""
+		s = s.substring(0, s.length()-1)
+	if paramCount > WC3LIMIT_FUNCTIONS_PARAMETER_COUNT
+		error(WC3LIMIT_FUNCTIONS_PARAMETER_COUNT_ERROR)
+	return s
+
+/** Creates one long String from all Strings delimited by comma.*/
+public function createDelimitedString(vararg string ids) returns string
```
Frotty: why is this an extra function? isn't it the same as `joinBy(",")` ?

wurst/util/StringUtils.wurst
```diff
@@ -157,6 +126,101 @@ public function string.reduce(string reduced) returns string
...
+		else
+			isDigit = false
+			return isDigit
+	return isDigit
+
+/** Checks if this string contains all characters in given string.
+	Can be useful for player commands as you can capture entered strings with typos.
+	If this string has different length than given string, it immediately returns false.
+	Example: "-etrne".containsAllAndOnly("-enter") == true */
+public function string.containsAllAndOnly(string chars) returns bool
```
Frotty: seems a bit esoteric to me. If you want to detect typos, you should probably just use a string distance function.

wurst/util/StringUtils.wurst
```diff
@@ -157,6 +126,101 @@ public function string.reduce(string reduced) returns string
...
+		for char in chars
+			if not this.contains(char)
+				containsAll = false
+				break
+	else
+		containsAll = false
+	return containsAll
+
+/** Checks whether a string is valid in terms of max WC3 string length for various usages.*/
+public function string.isValidLength() returns bool
```
Frotty: Specify `various` usages, or rather, rename to something properly, like `isValidDescription`. This is too generic.

wurst/util/StringUtils.wurst
```diff
@@ -157,6 +126,101 @@ public function string.reduce(string reduced) returns string
...
+	var isValid = true
+	if this.length() > WC3LIMIT_OBJECTS_DESCRIPTION_LENGTH
+		isValid = false
+		error("String |||"+this+"||| is too long. Length: "+this.length().toString())
```
Frotty: Use something sensible like `<>` instead of `||| |||`

wurst/util/StringUtils.wurst
```diff
@@ -157,6 +126,101 @@ public function string.reduce(string reduced) returns string
...
+			s += id
+			s += ","
+	if s != ""
+		s = s.substring(0, s.length()-1)
+	if paramCount > WC3LIMIT_FUNCTIONS_PARAMETER_COUNT
+		error(WC3LIMIT_FUNCTIONS_PARAMETER_COUNT_ERROR)
+	return s
+
+/** Checks whether a string is representing a digit or not.*/
+public function string.isDigit() returns bool
```
Frotty: The name is wrong, `digit` refers to a single element of a number, see definition 

> digit
/ˈdɪdʒɪt/
noun
any of the numerals from 0 to 9, especially when forming part of a number.

Also I think it would be easier to check if it's the string "0", and if not, use .toInt/.toReal which return 0 if it's not a valid number.

wurst/data/LinkedList.wurst
Donach: I just moved this function from StringUtils, mostly to just keep dependencies from failing... I did not create this function, it was there for some time already :-).

wurst/objediting/ObjectIds.wurst
Donach: Since its not public import, how can that break things? Also, the import was not used at all (maybe some functions which I actually moved from StringUtils to LinkedList were used by the ObjectIds, which was rendering this import obsolete)
Also, if I kept the import, there would be circular dependency problem.

wurst/util/StringUtils.wurst
Donach: Sure, will remove it if its hardcoded there.

wurst/util/StringUtils.wurst
Donach: Will do.

wurst/util/StringUtils.wurst
Donach: From what I've just tested, that does not have to be the case always. If you mix numbers and letters, it can convert part of those numbers from string to integer/real.
Therefore this function is safer, just needs a few more modifications.

wurst/util/StringUtils.wurst
Donach: joinBy is used to convert linkedlist to delimited string. This is creating delimited string from bunch of strings.

wurst/util/StringUtils.wurst
Donach: Removed, its probably too specific usage we have.

wurst/data/LinkedList.wurst
Krypt0n: the getals was not existing i think once we created it ^^

Donach:
Anything else to be added/changed/removed?

wurst/data/LinkedList.wurst
Frotty: You shouldn't mark stuff as resolved which isn't. I still see no point in this - either present a reason or remove.

wurst/util/StringUtils.wurst
```diff
@@ -17,6 +18,7 @@ int array width
 constant MAX_INDEX = '~' + 1
 constant DEFAULT_WIDTH = 0
 constant DEFAULT_STRING = ""
+constant let	 WC3LIMIT_OBJECTS_DESCRIPTION_LENGTH = 1023 // Should be a global for whole WurstScript or StdLib at least
```
Frotty: why is there a tab after "let" ?

wurst/util/StringUtils.wurst
```diff
@@ -157,6 +124,84 @@ public function string.reduce(string reduced) returns string
...
+		while pos != -1
+			cb.callback(this.substring(i, pos))
+			pos += 1
+			i = pos
+			pos = this.indexOf(delim, pos)
+			if pos == -1
+				cb.callback(this.substring(i, this.length()))
+
+/** Creates list of IDs converted from Int to their String value delimited by comma (using toRawCode).*/
+public function createDelimitedString(vararg int ids) returns string
```
Frotty: Where is the different to commaList ?

wurst/util/StringUtils.wurst
Frotty: Yea sry, I think I ment `commaList`.

Frotty:
Yup, I'm a bit busy recently. A few things still seem like duplicates. If I'm wrong, please add details explaining what to use this for/why not use existing func.

wurst/util/StringUtils.wurst
Donach: oh cause when I had the other constants, I tabbed them to start at same place... Nevermind, gonna change it.

wurst/util/StringUtils.wurst
Donach: That it does not use LinkedList inside the processing. Not really sure if its faster or anything, but the less using of classes the better I'd say.

wurst/data/LinkedList.wurst
Donach: Removed.

Krypt0n: well Linkedlist only has:
.peek()
and 
LinkedList.backiterator().next()
while peek does not truely provide the same functionality

wurst/util/StringUtils.wurst
Krypt0n: yeah at some point you have to many instantiated linked lists, when you canont have more then max jass array size lists at the same time but you maybe need 1 million

wurst/data/LinkedList.wurst
Frotty: `getFirst` and `peek` return last/first added elements

wurst/util/StringUtils.wurst
Frotty: > Not really sure if its faster or anything, but the less using of classes the better I'd say.

I wouldn't say that.

> yeah at some point you have to many instantiated linked lists, when you canont have more then max jass array size lists at the same time but you maybe need 1 million

That sounds more like an issue with your map architecture and not common use case imo. Additionally with the lua backend, those limits probably change.

Donach: I would not limit it to "map architecture" problems, just that the bigger the map gets, the higher the requirements :)
Anyway, gonna remove this now, does not need to be in stdlib.

Donach: Removed in 16e11dd

Donach:
Removed unnecessary functions
Anything else needed? If not, this can be merged @Frotty @Trokkin .

Thanks.

Frotty:
Im not a fan of moving joinBy functions into LinkedList, but in theory yes.

PR #245, opened 2019-05-12
Added new Trigger events, created Trigger class

Donach:
Created new Trigger class, which contains all getters and setters for events
EDIT: Just to explain, this was done and I thinks its useful due to fact Blizz adds (and may add more) getters which have the "Blz" prefix, its getting messy to grab trigger variables properly.
The naming might be still altered a bit ofc.


Donach:
Yeah that is open for discussion I guess
I like:
Trigger
TriggerData

To me, simply "Trigger" seems logical, as its just a "wrapper" for all functions refedencing the "trigger that fired"...
Anyway, add your opinions ideas @Frotty

Krypt0n:
Well yeah but I think `Trigger` is bad as it is very close to the native datatype `trigger`
 while your `Trigger `class is static function wrapper and not a wrappter of `trigger`

Frotty:
Why are there so many unrelated changes?

I guess the main benefit of this is to improve auto complete?

Krypt0n:
That is one thing. But the other point is that you have many different native getter functions with different namings and thisway you have all at one place.
Not really needed just a little easier sometimes.

Donach:
Yeah, and reason why added it here instead of master is to include also the new natives...

Frotty:
Fix the merge conflict please and remove the `SETTERS` and  `GETTERS` comments.
And I think `TriggerData` is more like `EventData` and belongs into the `EventHelper` package, since there we already have similar functions which could be soft replaced.

Donach:
Hey @Frotty, can you resolve the last conflict? I just added newline at the end of the tests file.

Besides that, should be ready for merge.

Frotty:
Yeah, I fixed conflict and some other stuff.
Looking good, will merge 👍 
Now we only need to change the documentation pages 😅

Frotty:
Thanks 🍰

wurst/event/EventHelper.wurst
```diff
@@ -69,3 +73,321 @@ public function eventid.isKeyboardEvent() returns boolean
 public function eventid.isMouseEvent() returns boolean
 	let eventId = this.getHandleId()
 	return (eventId >= 305 and eventId <= 307)
+
+/** This class exposes all Getters and Setters for trigger events.
+	Note: As these functions simply wrap natives, they will only return proper values during event callbacks (actions/conditions).
+	To use these values in a subsequent closure, save the event data in a variable at the time the event was fired. */
+public class EventData
```
Frotty: @Krypt0n @Donach I just noticed this now..
Is there any reason you went totally wild with the function names?
Often the naming is completely inverse of the native (just confusing ppl)
or it's even lacking original information, e.g. `getDialogButton` vs `GetClickedButton`.
I'm thinking about changing them on master as long as most people haven't adapted yet..

Krypt0n: "Is this stricly for autocomplete? or what"  Yes
- The idea was to have some consistency for trigger getters and namings, and also the hotdoc should include from which triggers events the getters can be used. Also one use case for this commit was (I think its now 6 months ago once we first talked about adding these things) to keep the inversed pattern logic from jass to wurst. Thus having more buttom-up and less top-down syntax. Its a bit of a point if you prefer predictive parsing or resursive descent parsing in the end.

I had no time yet to check the commit there. I will catch it up soon, and then come back with some proper namings and hotdoc.

And I agree, this does not seem to be of any use right now tbh ..

PR #246, opened 2019-05-13
Ptr 1.31 frameapi

Krypt0n:
added asset class for framehandelnames; updated framhandle hotdoc


Frotty:
You are trying to merge into master, I don't think that's correct?

Krypt0n:
Should be change several functions namings? I really don't like `.size` as a setter function

Frotty:
> Should we change several functions namings? I really don't like .size as a setter function

Yea probably since that's in line with data structures.

Looking good 👍

Krypt0n:
Okay will update then, so no need to merge by now :)

Krypt0n:
What do you prefer, something like this:
Getter = framehandle.size()
Setter = framhandle.setSize(real x, real y)
?

Frotty:
I think `.size()` should only be used, if size is readonly, i.e. no setter exists.
If setter and getter exist, you should use setXXX getXXX as usual.

Krypt0n:
Fixed namings, next point will be handling of syncronizing, cause stuff like editbox checkboxes etc. like to desync
so i might work on something with these:'

```
native BlzGetTriggerSyncPrefix                     takes nothing returns string
native BlzGetTriggerSyncData                       takes nothing returns string
native BlzSendSyncData                             takes string prefix, string data returns boolean
native BlzTriggerRegisterPlayerSyncEvent           takes trigger whichTrigger, player whichPlayer, string prefix, boolean fromServer returns event
```

Frotty:
But can merge before that ?

Krypt0n:
yes can be merge, these topics can be added to hotdoc or with some other handles later

Frotty:
Alright, thanks again 👍

PR #249, opened 2019-05-24
ClosureForGroups fix

WareditorGit:
Changed the new forUnitsInRange() to allow for the () -> syntax.
I believe the callback has to be the last argument.


Frotty:
Ah yes, seems we missed that. 👍

PR #252, opened 2019-05-30
Damage Event for 1.31

WareditorGit:
DamageEvent now supports new 1.31 natives:
- Unreduced listeners (damage event that fires before the damage is reduced by any damage reducing effect such as armor)
- attacktype, damagetype and weapontype handling
- Better and easier native spell detection (It is now detected by checking if the attacktype is ATTACK_TYPE_NORMAL)

Some things to consider:
- A way to deprecated EventListener.add(EVENT_UNIT_DAMAGED) is needed since we have a general event now
- If you set damage to zero or less during the unreduced phase, the reduced events will not fire (that's how it works natively).

### DIFF
--- a/wurst/event/DamageEvent.wurst
+++ b/wurst/event/DamageEvent.wurst
@@ -4,14 +4,10 @@
     **IMPORTANT** 
         This package **doesn't** provide protection from cyclic damage.
         Damage from code **must** be declared by the user using DamageEvent.setNextDamageFromCode() before dealing damage.
-        
+        If you set damage amount to zero or less in the unreduced listeners, reduced listeners will not fire.
+
         if DETECT_NATIVE_ABILITIES is true:
-            > Spell damage is converted into negative damage values without this system, so do not use any other damage detection method.
-            > Spell reduction abilities like Runed bracers or Elunes Grace have to be coded.
-            > Life Drain doesn't heal the caster.
-            > Locust Swarm needs its Damage Return Factor changed from 0.75 to -0.75.
-            > Mana Shield abilities need their Damage Absorbed % reduced to 0.00 (and the damage absorbtion part has to be coded)
-            > Anti-Magic Shell (amount-based) has to be coded.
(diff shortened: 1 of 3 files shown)


Frotty:
Nice, thanks 👍 
I'm unsure whether `unreduced` is the best naming here, but I couldn't come up with something better immediately.

Regarding deprecation, I suppose it would require referencing a variable inside the annotation, probably a new one. We should probably create an additional ticket for that.

wurst/event/DamageEvent.wurst
```diff
@@ -58,34 +60,18 @@
...
-    Also keep in mind if true:
-        > Spell damage is converted into negative damage values.
-        > Spell reduction abilities have to be coded.
-        > Life Drain doesn't heal the caster.
-        > Locust Swarm's Damage Return Factor has to be changed from 0.75 to -0.75.
-        > Mana Shield abilities' Damage Absorbed % has to be reduced to 0.00
-        > Mana Shield's damage absorbtion part & Anti-Magic Shell (amount-based) have to be coded.
-    If false, native abilities that deal damage are passed with DamageType ATTACK */
-@configurable constant DETECT_NATIVE_ABILITIES = false
+/** If true, any  damage instance of attacktype ATTACK_TYPE_NORMAL is passed with DamageType SPELL */
```
Frotty: double space here

wurst/event/DamageEvent.wurst
```diff
@@ -126,34 +112,55 @@ public class DamageElement
...
+        this.nativeWeaponType = nativeWeaponType
         this.damageType = damageType
         this.damageElement = damageElement
 
         count++ 
         stack[count] = this 
         current = this 
 
+    protected function setAmount(real amount)
+        if unreduced
```
Frotty: Could this be simplified to 
```
this.amount = amount
if unreduced
    this.unreducedAmount = amount 
```

Frotty:
I guess unreduced it is 👍 
Thanks

PR #256, opened 2019-06-09
Update TimedLoop module

WareditorGit:
Added the possibility to have different periods between different module implementations by overriding the new getPeriod() function.

### DIFF
--- a/wurst/util/TimerUtils.wurst
+++ b/wurst/util/TimerUtils.wurst
@@ -94,8 +94,8 @@ public enum TimedLoopState
 
 
 public module TimedLoop
-	static private thistype array instances		// The array
-	static private int instanceCount = 0	// The count
+	static private thistype array instances
+	static private int instanceCount = 0
 				
 	static private TimedLoopTimer timedLoopTimer 
 	static private triggercondition condition 
--- a/wurst/util/TimerUtils.wurst
+++ b/wurst/util/TimerUtils.wurst
@@ -100,7 +100,7 @@ public module TimedLoop
 	static private TimedLoopTimer timedLoopTimer 
 	static private triggercondition condition 
 
-	protected TimedLoopState state = PAUSED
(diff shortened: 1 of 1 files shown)


Frotty:
I think the multiple timers kind of get rid of the performance benefit?
Maybe a timestamp could be preferred here?

WareditorGit:
Maybe keep the original TimedLoop and add another module with different period per implementation possible?

One other idea would be to have timers associated to their period and when a class request a timer of a given period it would check if that timer already exist and run and would add the class function to the list of the callbacks of that timer. Might be overkill.

WareditorGit:
I went with the second idea. Tell me what you think.

Frotty:
Should be fine this way I suppose.
@Cokemonkey11 what do you think?

Cokemonkey11:
I don't use this package, so:

- what's this for?
- what's the package for?
- where's the docs?
- why would you want to override the period?

WareditorGit:
I didn't create the package, I just made it so the TimedLoop module could support different periods in different class implementation.

Cokemonkey11:
The diff for that should be just a few lines, no?

- Not saying this can't be merged as-is, but
- I'm not the best person for that

wurst/util/TimerUtils.wurst
```diff
@@ -45,75 +46,110 @@ public function timer.release()
...
+/* TimedLoop for classes */
 
-function evaluateTimedLoopTrig()
-	TriggerEvaluate(timedLoopTrig)
-	if conditionCount == 0
-		timedLoopTimer.pause()
+class TimedLoopTimer 
 
-enum Mode
+	constant timer loopTimer = getTimer()..setData(this castTo int)
```
Frotty: should omit types here

Frotty:
Thanks 👍

PR #257, opened 2019-06-11
Update DamageEvent

WareditorGit:
It used to be if your damage was 0 or less in the pre-reduction phase
it did not trigger the normal on damage event ( EVENT_PLAYER_UNIT_DAMAGED ).
In 1.31.1 it does, hence the need for an update.


Frotty:
:+1:

PR #260, opened 2019-06-16
add animations by index to Assets package

Cokemonkey11:
Thanks @WareditorGit for source material

Warning 20k+ lines (don't load in Github)


Frotty:
What exactly are you using this for? 🍡 
Could be useful I guess. Do you want to generate some other missing asset packages too?

Cokemonkey11:
> What exactly are you using this for?

It's useful for playing a unit's animation

> Could be useful I guess. Do you want to generate some other missing asset packages too?

I didn't generate this - someone else provided a text dump when I asked if anyone had a mapping from name to index.

I've just transformed that into wurst.

Do you have other dumps that are potentially useful?

This one in particular is immediately valuable for me.

PR #261, opened 2019-06-18
Add reset method to HLIterator

frederikaalund:
Small addition. Should be backwards compatible.

I need this to implement an efficient `staticItr()` in a module.


Frotty:
seems fine 👍

PR #262, opened 2019-06-22
fixed some errors encountered while testing with Lua

peq:
- multiboars in lua do not accept null values, so added a workaround
- lua checks member access, fixed a related NPE in ClosureTimers


Frotty:
I hope the closureEvents change doesn't break anything :D
Thanks 👍

PR #268, opened 2019-07-12
LinkedListIndexed

WareditorGit:
A module that turns a class into a collection of linked lists where each list is composed of instances of the same index.

This will be useful for closureFrames and closureKeyPresses.

Not sure about the name.


Frotty:
> This will be useful for closureFrames and closureKeyPresses.

It should also be able to be used in ClosureEvents right?

WareditorGit:
Yeah and others if needed

wurst/data/LinkedListIndexedModule.wurst
```diff
@@ -0,0 +1,130 @@
+package LinkedListIndexedModule
+import HashMap
+
+/** Turns a class into a collection of linked lists where each list
+    is composed of instances of the same index. Each instance 
+    knows it's previous and next member in its index's list. */
```
Cokemonkey11: Doesn't answer "Why?". What's this package useful for?

wurst/data/LinkedListIndexedModule.wurst
```diff
@@ -0,0 +1,130 @@
...
+        function next() returns LinkedListIndexedModule.thistype
+            let res = current
+            current = current.prev
+            return res
+
+        function reset()
+            current = last
+
+        function close()
+            destroy this
```
Cokemonkey11: missing eof lf

wurst/data/LinkedListIndexedModule.wurst
Frotty: To modularize stuff we are already doing in the stdlib.

Frotty:
@WareditorGit can you fix eof lf? should adhere to editorconfig.
Did you have any further ideas about other names?

WareditorGit:
Sorry, I am away from my comp during this period. Will work on this when I get back.

Frotty:
@WareditorGit I could also merge like this or are there changes u want to make besides name?

WareditorGit:
No changes.
However I agree that it's might be too peculiar to include in the std lib and might add some unrequired abstraction.

Frotty:
What's the point of not having it in stdlib if we plan to use it in stdlib?

WareditorGit:
I am saying maybe not use it if it's only for those few cases.

Frotty:
Will this work for destroying listeners during a callback without temp var?
 I still want to merge tbh.

Frotty:
bump

Cokemonkey11:
> What's the point of not having it in stdlib if we plan to use it in stdlib?

If you plan to use in stdlib then it should be in stdlib.

It would be interesting to see a concrete stdlib change - in this PR - that moves old code to using this new feature.

Frotty:
I actually like this, but the hashmaps seem overkill, why not just use module attribute. Might fix and merge.

PR #269, opened 2019-07-24
Allow dummy caster unit to attack when ordered

Rhazarian:
This change allows to use dummy casters with "attackonce" order. It may be useful in the following pattern:
`new DummyCaster().castTarget(AbilityIds.poisonSting, 1, OrderIds.attackonce, target)`
The target unit will be poisoned.

This does not break compatibility since the dummy won't attack unless explicitly ordered thanks to the special ability based on permanent invisibility.

### DIFF
--- a/wurst/dummy/DummyRecycler.wurst
+++ b/wurst/dummy/DummyRecycler.wurst
@@ -186,7 +186,7 @@ init
 	..setAttack1DamageSidesperDie(1)
 	..setAttack1ProjectileSpeed(INT_MAX)
 	..setAttack1Range(INT_MAX)
-	..setAttack1TargetsAllowed("air,debris,ground,invulnerable,item,structure,vulnerable,ward")
+	..setAttack1TargetsAllowed(commaList(TargetsAllowed.air, TargetsAllowed.debris, TargetsAllowed.ground, TargetsAllowed.invulnerable, TargetsAllowed.item_t, TargetsAllowed.structure, TargetsAllowed.vulnerable, TargetsAllowed.ward))
 	..setAttack1WeaponSound(WeaponSound.Nothing)
 	..setAttack1WeaponType(WeaponType.Instant)
 	..setAcquisitionRange(INT_MAX.toReal())
--- a/wurst/dummy/DummyRecycler.wurst
+++ b/wurst/dummy/DummyRecycler.wurst
@@ -150,6 +150,7 @@ public constant DISABLE_AUTO_ACQUIRE_ATTACK_TARGETS_ABILITY_ID = 'xdaa'
 	..presetAutoAcquireAttackTargets(_ -> false)
 	..presetDurationNormal(_ -> -1)
 	..presetDurationHero(_ -> -1)
+	
 	new UnitDefinition(DUMMY_UNIT_ID, UnitIds.wisp)
 	..setName("Effect Dummy Unit")
 	..setUpgradesUsed("")
--- a/wurst/dummy/DummyRecycler.wurst
+++ b/wurst/dummy/DummyRecycler.wurst
@@ -7,7 +7,6 @@ import ErrorHandling
 import MapBounds
 import GameTimer
 import AbilityObjEditing
-import initlater ObjectIdGenerator
 import ObjectIds
 import Colors
 import Assets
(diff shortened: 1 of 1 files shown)


wurst/dummy/DummyRecycler.wurst
```diff
@@ -170,5 +176,19 @@ init
...
+	..setAttack1AnimationBackswingPoint(0)
+	..setAttack1AnimationDamagePoint(0)
+	..setAttack1CooldownTime(0)
+	..setAttack1AttackType(AttackType.Unknown)
+	..setAttack1DamageBase(-1)
+	..setAttack1DamageNumberofDice(1)
+	..setAttack1DamageSidesperDie(1)
+	..setAttack1ProjectileSpeed(INT_MAX)
+	..setAttack1Range(INT_MAX)
+	..setAttack1TargetsAllowed("air,debris,ground,invulnerable,item,structure,vulnerable,ward")
```
Frotty: you replaced the others, so should use `TargetsAllowed` package here too 😀

Frotty:
Hi,
thanks for the contribution 👍 
I'm not a big fan of `initlater` imports in the standardlib. Can this be avoided?
Otherwise looking good.

@Cokemonkey11 ?

Rhazarian:
Not sure how to avoid `initlater`.
`DummyRecycler` is used in `Preloader`, which is used in `ObjectIdGenerator` to preload generated abilities.
`ObjectIdGenerator` is used here to generate an id for the special ability.
I may hardcode an id for this ability, but is this a good solution, or are there any other ideas?

wurst/dummy/DummyRecycler.wurst
```diff
@@ -143,6 +145,10 @@ init
 			DummyRecycler.angleQueues[i].enqueue(createDummy(boundMax, DUMMY_PLAYER, (i* ANGLE_DEGREE).asAngleDegrees()))
 
 @compiletime function generateDummyUnit()
+	let a = new AbilityDefinitionPermanentInvisibility(ABIL_ID_GEN.next())
+	a..presetAutoAcquireAttackTargets(_ -> false)
```
Cokemonkey11: I think there's no need for a `let` binding here. Just `new X()` followed by cascades.

wurst/dummy/DummyRecycler.wurst
```diff
@@ -143,6 +145,10 @@ init
 			DummyRecycler.angleQueues[i].enqueue(createDummy(boundMax, DUMMY_PLAYER, (i* ANGLE_DEGREE).asAngleDegrees()))
 
 @compiletime function generateDummyUnit()
+	let a = new AbilityDefinitionPermanentInvisibility(ABIL_ID_GEN.next())
```
Cokemonkey11: Is there some documentation for this behavior somewhere, in particular the `-1` values? Given blizzard's recent changes, I'm just a little wary of relying on undefined behaviour.

wurst/dummy/DummyRecycler.wurst
Rhazarian: I agree. There was a `let` binding in the previous version (on the creation of UnitDefinition), so I decided not to remove it since it was in stdlib already and to do the rest in the same style. Gonna refactor it now.

wurst/dummy/DummyRecycler.wurst
Rhazarian: I have tested it on the previous versions and on the current one, and it works. I have found this 'hack' a few years ago and I have been using it since then, could not find it now though. There is another way to achieve the same effect, but it will require changing acquire range after the unit creation using a native, since changing AR to 0 at compile time will reduce unit's attack range. It could be a fine solution if the dummy is used only through the recycler API, but it's ID is visible to everyone so it could be created in other places and then it's gonna bug.

Rhazarian: While I'm a little bit wary too, I think that the chance of Blizzard suddenly changing the effect of a slightly 'nonsence' value used in a single one, not frequently used and not causing any troubles ability is extremely low.

wurst/dummy/DummyRecycler.wurst
```diff
@@ -143,14 +142,16 @@ init
...
 @compiletime function generateDummyUnit()
-	let a = new AbilityDefinitionPermanentInvisibility(ABIL_ID_GEN.next())
-	a..presetAutoAcquireAttackTargets(_ -> false)
+	new AbilityDefinitionPermanentInvisibility(DISABLE_AUTO_ACQUIRE_ATTACK_TARGETS_ABILITY_ID)
+	..presetAutoAcquireAttackTargets(_ -> false)
 	..presetDurationNormal(_ -> -1)
 	..presetDurationHero(_ -> -1)
-	let u = new UnitDefinition(DUMMY_UNIT_ID, UnitIds.wisp)
-	u..setName("Effect Dummy Unit")
+	new UnitDefinition(DUMMY_UNIT_ID, UnitIds.wisp)
```
Cokemonkey11: Maybe linebreak for readability?

wurst/dummy/DummyRecycler.wurst
Cokemonkey11: Fine with me.

I would like to see thinking like this described in comments, but the quality there across the standard library is pretty variable, so I am not too fussed.

Frotty:
Not a big fan of random constant id, but I guess it's fine since we do it for the dummy as well?

Cokemonkey11:
I've just remembered one concern I had with using constants - isn't it possible for the ID_GEN tools to create a duplicate?

Rhazarian:
It is possible, though not highly likely, since hundreds of ids should be generated because they're generated in alphabetical order.

We have to choose the lesser of the two evils - either using `initlater` import or a hardcoded constant, or probably refactor preloader and idgen packages somehow, so that the `initlater` import would not be required, but it may require a lot of effort and i'm not 100% sure that it is possible to refactor it in the wanted way.

Cokemonkey11:
I'm happy that `xdaa` won't get autogenerated unless you have a large number of objects

Frotty:
Looking good. One last thing,
could you add the 

> `new DummyCaster().castTarget(AbilityIds.poisonSting, 1, OrderIds.attackonce, target)`
The target unit will be poisoned.

fact and example somewhere? Ideally on stdlib doc page ❓

Rhazarian:
Done.

Cokemonkey11:
Looks great!

Frotty:
Cool, thanks a lot! 👍

PR #270, opened 2019-07-25
Fix inifinite recursion in unit.isAllyOf

frederikaalund:
The bug causes the map to freeze.


frederikaalund:
Side note: Is the `this != null` test really needed? Very few native wrappers from the standard library does this. Is `IsAllyOf` special somehow?

Frotty:
Hi thanks for the fix 👍 

> Side note: Is the this != null test really needed? Very few native wrappers from the standard library does this. Is IsAllyOf special somehow?

Idk, does the native return true or false for null units?

PR #271, opened 2019-07-25
Make reflection natives public

frederikaalund:
Otherwise, they cannot be used by others.


Frotty:
ouch 😅 thanks for the fix 👍

PR #277, opened 2019-08-02
Fixup animation duplicates

Cokemonkey11:
Now that @peq has added support for static inner classes, the compiler is flagging some duplicate animation names.

This PR fixes that up.

The second commit replaces the constant members with static constant, as requested.

// CC @frederikaalund FYI


Frotty:
🍭

PR #278, opened 2019-08-03
Update ClosureEvents.wurst

rolandc85:
Fix event destroyed during callback to break the loop of calling other events


Frotty:
Cool, thanks for your first contribution 🎉 🍰 🍻 👍

PR #280, opened 2019-08-04
Remove UnitAnimations package

Cokemonkey11:
Removing this package from stdlib because it caused multiple issues, most recently wurstscript/WurstScript#883.

Use https://github.com/Cokemonkey11/WurstUnitAnimations instead.


Frotty:
I don't think removing it completely makes much sense, especially since people are already using it.

Cokemonkey11:
// cc @frederikaalund 

I feel somewhat strongly that this package causes compile-time slowness and is more trouble than it's worth.

The new dependency management isn't perfect, but it's very much good enough for this use-case IMO

frederikaalund:
I vote to keep `UnitAnimations` in the standard library.

Whether to include a package in the standard library or have it available through a package manager is always a hot topic for any language (see [python](https://lwn.net/Articles/776239/) for instance). Either way has its set of pros and cons.

I'd like to have the compiler changes required for `UnitAnimations` since said changes will welcome additional libraries in the same vein. I think that would strengthen Wurst as a whole.

Furthermore, I think it is nice to see `UnitAnimations` push the edge of what Wurst can do. With `UnitAnimations` in the standard library, we are guaranteed support for the necessary compiler features even in future versions of Wurst. More libraries like this in the standard library, please. I hope that @peq and @frotty can keep up with the maintenance burden. :)

Frotty:
So my suggestion would be to just remove the package from the Assets package, thus not getting automatically imported unless used.

Cokemonkey11:
I'm ok with that

Frotty:
I did the thing 🍰

PR #283, opened 2019-08-08
Add splice method for linked list

Rhazarian:
Allows to transfer elements from a list in a constant time.

### DIFF
--- a/wurst/data/LinkedList.wurst
+++ b/wurst/data/LinkedList.wurst
@@ -22,19 +22,19 @@ public class LinkedList<T>
 	private LLIterator<T> staticItr = null
 	private LLBackIterator<T> staticBackItr = null
 
-	/** Create a new list by copying all elements from another list into it */
+	/** Creates a new list by copying all elements from another list into it */
 	construct(thistype base)
 		dummy.next = dummy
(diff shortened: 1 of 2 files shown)


Frotty:
Cool, could be useful 👍 . But shouldn't the parameter list be destroyed? This is more like a addAll and clear, except not destroying elements.

Rhazarian:
The idea and the name were taken from C++.

https://en.cppreference.com/w/cpp/container/list/splice

Yes, it does the same job as sequenced `addAll` and `clear`, but a lot more faster.

I'm not sure about the destruction of the parameter. On the one hand, it could be useful by allowing you to pass a list returned from another function directly to this one. On the other hand, the parameter list may be wanted to still be used somehow, and implicit destruction may be confusing.

Frotty:
Okay, I guess it's fine then.
Can you add a similar hint then, like in the doc you linked, that this doesn't remove/add elements (relevant for derived lists e.g.) and that the param list will be empty but not destroyed.

Rhazarian:
Extended the hotdoc.

Frotty:
nice, thanks 👍

PR #285, opened 2019-08-11
new bitset

Rhazarian:
Bitset is now superfast - only one native call (and nothing more) per operation.
Powers of 2 are now precomputed and new bitwise natives are used instead of some mathematical magic.
Added `flip` method, which allows to change the state of a bit to the contrary one.
Added new tests.


Frotty:
LGTM

wurst/data/BitSet.wurst
```diff
@@ -1,31 +1,46 @@
...
-	return not this.containsPow(pow) ? bitset(this.val + pow) : this
+	return bitset(BlzBitOr(this.val, pows[v]))
 
 /** removes value v from the bitset */
 public function bitset.remove(int v) returns bitset
-	let pow = 2 .pow(v)
-	return this.containsPow(pow) ? bitset(this.val - pow) : this
+	return bitset(BlzBitAnd(this.val, reversePows[v]))
+
+/** adds value v to the bitset if the bitset does not contain it, removes it otherwise */
```
Cokemonkey11: What domain of v is acceptable?

wurst/data/BitSet.wurst
```diff
@@ -47,3 +62,28 @@ public function bitset.remove(int v) returns bitset
 
 @test function testRemove()
 	bitset(45).remove(3).val.assertEquals(37)
+
+@test function testFlip()
+	emptyBitset().add(3).flip(3).contains(3).assertFalse()
+	emptyBitset().flip(3).contains(3).assertTrue()
```
Cokemonkey11: Maybe add some tests for boundary conditions

Rhazarian: I've added a test.

wurst/data/BitSet.wurst
Rhazarian: It is now stated by the constant above.

Rhazarian:
I have redesigned the API to make it more modern. It's now more similar to Java & C++ bitsets.
Old method names are kept for compatibility.

Frotty:
Cool, is it done yet?

> Old method names are kept for compatibility.

If they duplicate functionality, you should deprecate them.

Rhazarian:
Yea, I guess it's done.

Frotty:
Yo, by deprecating old functions I means adding `@Deprecated("use func x instead").
NOT deleting them...
That just breaks everything.

Rhazarian:
I did exactly that?

Frotty:
Ah okay, I searched for `contains(int v)` but there was only a deletion match since you renamed the parameter..
Alright, thanks 👍

PR #286, opened 2019-08-12
add root ability when create dummy

yefq:
Make dummy not need to turn when spell or attack.
This will allow dummy to cast spell on countless targets at the same time without missing.


Cokemonkey11:
LGTM

Frotty:
Yeah, thanks again @yefq 👍 
The only issue is that dummies can't move anymore, but I think that shouldn't be expected in the first place.

yefq:
@Cokemonkey11 @Frotty 
:heart::heart::heart:

Frotty:
Did you see my edits?
Does this cause any issues with attacking?

yefq:
> Did you see my edits?
> Does this cause any issues with attacking?

I ran a simple test, there is no problem.
Attack not need turn, and orb effect is work normal.
(Test with Orb `'Apoi'` and `'ACvs'`)

Frotty:
👍 🍰

yefq:
> 👍 🍰

👍  🍰

PR #288, opened 2019-08-22
Fixing peanut color code

SSappy:
Color code found here https://www.hiveworkshop.com/threads/need-help-with-player-color-codes-and-where-to-find-them-all.314460/.
A 1 was missing. color(164, 11, 51) instead of color(164, 111, 51).


Frotty:
Thanks 🍬 👍

PR #290, opened 2019-09-01
fix ClosureForGroups.forNearestDestructable()

Rhazarian:
Previous version was not working it all. It was using filterEnumDestructablesInCircleBJ, which uses some bj vars and requires them to be set, which were not. It could be fixed with just setting this variables, but previous implementation has some other issues: distance was calculated two times (in the filter function and in the enum function) and the rect created was not removed! Now a global rect is reused, and the in-rect-check was moved to enum function to avoid recalculating distance.


Frotty:
What is this 100k line .j file doing in this PR ?
Otherwise looking good 👍

Rhazarian:
Holy cow, didn't even notice it was there 🤣

Frotty:
Okay, why did you remove the push/pop callback?
Now this doesn't work nested right?

Rhazarian:
It does work.
Other function use some global variables to pass the callback to the filter functions, so a stack is used there, because otherwise the global variable would be rewritten when a forDestructables...() func would be called inside a callback.
In this function callback is not passed somewhere.
I have forgot to destroy it though, it is fixed now.

Frotty:
Okay sorry, I didn't have time to test.
Are you vouching for this working correctly? 😅

Rhazarian:
Hey, sorry, didn't have time to check this PR c:
Regardless of this is already merged, I am going to answer:
Yes, I am using this in my map for a while now and it does work, unlike the previous version, which didn't work at all! :)

PR #293, opened 2019-09-14
Fix unexpected leave event

yefq:
Some unexpected trigger before this fix:
1. trigger five times when InstantDummy in use
(dummy created when map init, and locust can't be enum in rect, so it was without Defend ability)
2. new created unit issued order before add Defend ability
(auto order like Sorceress's SlowOn)

In short, unit may issued order when it without Defend ability,
an additional condition make this event more reliable.


Frotty:
Is that condition kosher? Or should the dummies be picked with a player-owned enum or so?

yefq:
Sorry, I fail to notice your reply.
I only added a simple condition, Because I want to be conservative.
After all, Dummies in the standard library are recycled, they don't usually leave the map.

PR #294, opened 2019-09-20
Multiboard ext. functions

GetLocalPlayer:


Frotty:
No description?
Looks good though 👍

Frotty:
I think unit tests would be nice, but I would need to implement the natives for compiletime first.

GetLocalPlayer:
Well, a bit better API for minimizing, showing and hiding.
```
multiboard.show()
multiboard.show(player)
multiboard.hide()
multiboard.hide(player)
multiboard.minimize()
multiboard.minimize(palyer)
multiboard.maximize()
multiboard.maximize(player)
```
Item functions to eliminate type conversion routine
```
multiboarditem.setValue(int)
multiboarditem.setValue(real)
multiboarditem.setValue(bool)
```
Particular item value setting directly through a multiboard instance
```
multiboard.setItemValue(row, column, value)
multiboard.setItemIcon(row, column, iconFileName)
```
So we don't have to repeat these annoying `.getItem`/`.release` all the time.
And more convenient width setting to a whole column or a row
```
multiboard.setRowWidth(row, width)
multiboard.setColumnWidth(column, width)
```

wurst/_handles/Multiboard.wurst
```diff
@@ -67,7 +101,51 @@ public function multiboard.setTitleColor(int r,int g, int b, int a)
...
+	..release()
+
+public function multiboard.setItemWidth(int row, int column, real width)
+	this.getItem(row, column)
+	..setWidth(width)
+	..release()
+
+public function multiboard.setColumnWidth(int column, real width)
+	if column >= this.getColumnCount()
+		return
```
Frotty: should probably show an error instead

GetLocalPlayer:
Would it be better to short `multiboard.setItemValue` functions to `multiboard.setValue`?

Frotty:
I think `itemValue` is fine, thanks 👍

PR #295, opened 2019-09-20
Fields are mixed up.

GetLocalPlayer:


GetLocalPlayer:
An Elune's Grace missing field has also been added.

Frotty:
hm, are there more issues like this?
These defs got generated and the "defx" might be wrong elsewhere?
Thanks for fix 👍

GetLocalPlayer:
A quick check doesn't show any other such issue in `AbilityObjEditing`. The only thing is `Def4` and `Def3` fields in Elune's Grace definition that are missing too. But these fields don't realy work for this ability. The only fields that actually work are `Def1`, `Def2` and `Def5`.

PR #301, opened 2019-10-08
Fix Dummy Facing

yefq:
Before this change:
1. wrong angle index when angle < 0 (always round down)
2. root ability was made dummy angle to 250 when it's created (default building facing)
3. createDummy was private function (but sometimes, we might want to create it for optimize efficiency in some libraries)


Frotty:
> createDummy was private function (but sometimes, we might want to create it for optimize efficiency in some libraries)

Why? You are supposed to use `.get` so you get a recycled one.
Otherwise looks good.

yefq:
> createDummy was private function (but sometimes, we might want to create it for optimize efficiency in some libraries)
> 
> Why? You are supposed to use .get so you get a recycled one.
> Otherwise looks good.

I think you are right ...

Frotty:
Thanks 👍

PR #303, opened 2019-10-17
add configurable heightmap provider

Cokemonkey11:
Enables Knockback3 to be used in environments where GetLocationZ can desync


Frotty:
👍

PR #304, opened 2019-10-20
New FileIO, Sync, SaveLoad

Frotty:
**Migration Notice:**
- `MultiFileIO`, `PreloadIO`, `IOTaskExecutor` : Use `FileIO` instead
- `Network`, `GamecacheBuffer`. `SyncSimple`: Directly sync strings using the new extension functions in `SyncSimple`
- `Persistable` will be replaced by `Serializable` in the future  

---

Replaces the current broken and legacy implementations of saving, loading and syncing data with new and improved ones using the newly added natives.
The main goals of these new packages is to be lightweight and have intuitive, simple API.

Future enhancements would be something like a `Serializable` interface to de/serialize classes, perhaps using the existing `OrderedStringBuffer` etc.

Usage examples are at the top of each package.


wurst/file/FileIO.wurst
```diff
@@ -0,0 +1,109 @@
+package FileIO
+import public ChunkedString
+import AbilityIds
+import ErrorHandling
+
+/* 	FileIO for Patches 1.31 and up with chunked string support for big data.
+	Reads and writes using preload files by manipulating default ability tooltips.
```
Cokemonkey11: How does this actual write and read from files with arbitrary filename?

wurst/file/FileIO.wurst
```diff
@@ -0,0 +1,109 @@
...
+
+	Read a file:
+	>	let fileContent = new File("MyFileName.pld").readAsString()
```
Cokemonkey11: This API example doesn't give me any reason to want a class at all - why not just `readToString(string filename)`?

wurst/file/FileIO.wurst
```diff
@@ -0,0 +1,109 @@
...
+
+	Write a file:
+	>	new File("MyFileName.pld")..write("MyContent")..close()
```
Cokemonkey11: Ditto here - why not just `writeToFile(string filename, string content)`?

wurst/file/FileIO.wurst
```diff
@@ -0,0 +1,109 @@
...
+*/
+
+/** Abilities which tooltips will be used to load data */
+@configurable constant IO_ABILITIES = [AbilityIds.aerialShackles, AbilityIds.rocketAttack, AbilityIds.callToArms,
```
Cokemonkey11: Why is this configurable?

wurst/file/FileIO.wurst
```diff
@@ -0,0 +1,109 @@
...
+
+	function close()
+		destroy this
+
+	function read() returns ChunkedString
+		return readPreload()
+
+	function readAsString() returns string
+		let read = readPreload()
+		return read != null ? read.getChunk(0) : "-"
```
Cokemonkey11: Why a hyphen?

wurst/file/FileIO.wurst
```diff
@@ -0,0 +1,109 @@
...
+		// Begin to generate the file
+		PreloadGenClear()
+		PreloadGenStart()
+
+		for i = 0 to buffer.chunkCount - 1
+			if (i >= IO_ABILITIES.length)
+				error("FileIO(" + filename + ") ERROR: String exceeds max length (" + I2S(IO_ABILITIES.length * buffer.chunkSize) + ").|r")
+
+			let chunk = buffer.getChunk(i)
+			Preload("\" )\ncall BlzSetAbilityTooltip(" + IO_ABILITIES[i].toString() + ", \"" + chunk + "\", 0)\n//")
```
Cokemonkey11: this wacky stuff needs comments

wurst/file/FileIO.wurst
```diff
@@ -0,0 +1,109 @@
...
+
+		Preload("\" )\nendfunction\nfunction a takes nothing returns nothing\n //")
+		PreloadGenEnd(this.filename)
+
+	private function readPreload() returns ChunkedString
+		if not CAN_READ_FILES
+			Log.warn("Reading files is not supported on this system.")
+			return null
+
+		string array original
```
Cokemonkey11: Why not precompute at map init?

wurst/file/FileIO.wurst
```diff
@@ -0,0 +1,109 @@
...
+	function write(string content)
+		let buffer = new ChunkedString()
+		buffer..append(content)..flush()
+		write(buffer)
+		destroy buffer
+
+	function write(ChunkedString buffer)
+		for i = 0 to buffer.chunkCount
+			let str = buffer.getChunk(i)
+			if validateInput(str) != null
```
Cokemonkey11: When there's a backslash or a quote you just log an error and plow on?

wurst/file/Persistable.wurst
```diff
@@ -1,220 +0,0 @@
...
-	Represents the various states of a persistable class load.
-	SUCCESS - The class has successfully loaded from disk.
-	FAIL_NOT_ENABLED - Local files have not been enabled for this user.
-	FAIL_FILE_CORRUPT - Decoding failed due to a corrupted/tampered file.
-	FAIL_DESERIALIZE - Decoding failed due to bad data in the file, meaning the .deserialize()
-	function errored out due to missing/malformed data after it has been succesfully decoded.
-	This can also happen if the file is empty or doesn't exist.
-	FAIL_NO_PLAYER - The player has left either during data transfer, or before.
-**/
-public enum LoadStatus
```
Cokemonkey11: Delete instead of deprecate?

wurst/file/PreloadIO.wurst
```diff
@@ -1,151 +0,0 @@
...
-	each up to MAX_PACKET_LENGTH characters.
-
-	WARNING 1: Due to the way how WC3 handles file IO, you have to use
-	this class only in one place at once. Avoid using it over time, lest you
-	risk causing conflicts with other clients of the class.
-	WARNING 2: WC3 does not correctly handle recursive folder creation.
-	If you are creating a file that is nested deeper than one folder, e.g.
-	"a/b/c", make sure to use IOWriter.createFolder("a/b/c") before that.
-**/
-public class IOWriter
```
Cokemonkey11: Delete instead of deprecate?

wurst/file/SaveLoadData.wurst
```diff
@@ -0,0 +1,82 @@
+package SaveLoadData
+import FileIO
+import SyncSimple
+import ErrorHandling
+
+/* 	Comfort wrapper for player based, synced access to save data.
+	For small data you can use strings, otherwise use ChunkedString.
```
Cokemonkey11: Seems weird to push the responsibility on the user

Can you support one or the other?

wurst/file/SaveLoadData.wurst
```diff
@@ -0,0 +1,82 @@
...
+
+	Save data for a player:
+	>	myPlayer.saveData("myFileName", "someDataString")
+	>	myPlayer.saveData("myChunkedFileName", someChunkedString)
+
+	Load data for a player:
+	>	myPlayer.loadData("myFileName") (status, data) ->
```
Cokemonkey11: I don't see any reason why this should consume a closure - why not just `let status_data = mPlayer.loadData("filename")`?

wurst/file/SaveLoadData.wurst
```diff
@@ -0,0 +1,82 @@
...
+			return "Failed! The player can't read files."
+
+public enum LoadStatus
+	SUCCESS
+	FAIL_PLAYER_OFFLINE
+	FAIL_FILE_EMPTY
+	FAIL_CANT_READ
+
+public function LoadStatus.isFail()	returns boolean
+	return (this castTo int) > 0
```
Cokemonkey11: I don't like this pattern, happy to discuss in more detail

Cokemonkey11: just use `this != LoadStatus.SUCCESS`?

wurst/file/SaveLoadData.wurst
```diff
@@ -0,0 +1,82 @@
...
+
+public interface LoadListener
+	function onLoad(LoadStatus status, ChunkedString data)
+
+/** Blockingly saves the given data string for the given player. */
```
Cokemonkey11: s/blocking/synchronous/g

wurst/file/SaveLoadData.wurst
```diff
@@ -0,0 +1,82 @@
...
+	this.saveData(slotName, buffer)
+	destroy buffer
+
+/** Blockingly saves the given data string for the given player. */
+public function player.saveData(string slotName, ChunkedString data)
+	if not this.isIngame()
+		error("Attempting to save data for player who isn't ingame! id: " + this.getId().toString())
+
+	if localPlayer == this
+		new File(slotName)..write(data)..close()
```
Cokemonkey11: Is this local-safe?

wurst/file/SaveLoadData.wurst
```diff
@@ -0,0 +1,82 @@
...
+public function player.saveData(string slotName, string data)
+	let buffer = new ChunkedString(DEFAULT_CHUNK_SIZE)
+	buffer.append(data)
+	this.saveData(slotName, buffer)
+	destroy buffer
+
+/** Blockingly saves the given data string for the given player. */
+public function player.saveData(string slotName, ChunkedString data)
+	if not this.isIngame()
+		error("Attempting to save data for player who isn't ingame! id: " + this.getId().toString())
```
Cokemonkey11: Seems good enough error handling imo

wurst/file/legacy/IOTaskExecutor.wurst
```diff
@@ -1,144 +0,0 @@
-package IOTaskExecutor
-import LinkedList
-import LinkedListModule
-import Execute
-import TimerUtils
-import ErrorHandling
-import GameTimer
-
-/** Simple interface of an IO task */
-public interface IOTask
```
Cokemonkey11: Delete instead of deprecate?

wurst/file/legacy/LocalFiles.wurst
```diff
@@ -1,38 +0,0 @@
-package LocalFiles
-import ClosureTimers
-import LinkedList
-
-interface LocalFilesCallback
```
Cokemonkey11: Delete instead of deprecate?

wurst/file/legacy/MultifileIO.wurst
```diff
@@ -1,253 +0,0 @@
-package MultifileIO
-import PreloadIO
-import ErrorHandling
-import Buffer
-import IOTaskExecutor
-
-class FilePacket
```
Cokemonkey11: Delete instead of deprecate?

wurst/network/GamecacheBuffer.wurst
```diff
@@ -1,128 +0,0 @@
...
-import GamecacheKeys
-import Execute
-import NetworkConfig
-import ErrorHandling
-import SafetyChecks
-
-interface SynchronizerFunction
-	function sync(string key)
-
-public enum BufferType
```
Cokemonkey11: Delete instead of deprecate?

wurst/network/GamecacheKeys.wurst
```diff
@@ -1,34 +0,0 @@
...
-import Execute
-import ErrorHandling
-import NetworkConfig
-
-// characters that can be used as GC keys
-// their amount also determines the size of a single 'round'
-constant GC_KEYS = " !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`{|}~"
-constant GC_KEYS_LENGTH = compiletime(GC_KEYS.length())
-
-public class GamecacheKeys
```
Cokemonkey11: Delete instead of deprecate?

wurst/network/Metadata.wurst
```diff
@@ -1,52 +0,0 @@
...
-
-public tuple metadata(int ints, int reals, int booleans, int asciiInts, int syncs)
-
-public constant EMPTY_META = metadata(0, 0, 0, 0, 0)
-
-/**
-	Simple wrapper class for Network-related metadata about the transmission.
-	Necessary to let other clients know about the amount of data being sent.
-**/
-public class MetadataStore
```
Cokemonkey11: Delete instead of deprecate?

wurst/network/Network.wurst
```diff
@@ -1,473 +0,0 @@
...
-	FINISHED
-
-public enum NetworkResult
-	SUCCESS
-	ABORTED
-
-interface NetworkFinishedCallback
-	function onFinish(NetworkResult status, HashBuffer buffer)
-
-public class Network
```
Cokemonkey11: Delete instead of deprecate?

wurst/network/SyncSimple.wurst
```diff
@@ -1,181 +1,99 @@
...
-		this.select(what)
-		this.unselect(what)
-
-	if count >= 12
-		this.select(last)
-
-public interface SynchronizationCallback
-	function onSynchronized()
-
-public class SimpleSynchronizer
```
Cokemonkey11: Delete instead of deprecate?

wurst/file/ChunkedString.wurst
```diff
@@ -0,0 +1,86 @@
...
+/*	A chunked string splits up big strings into chunks,
+	which is used to circumvent the maximum string length limit
+	and to prepare data for chunked storage, e.g. for FileIO.
+	Make sure to flush the chunked string when done writing to it,
+	so the unfilled buffer gets written as well.
+
+	Write Data:
+	>	let cstring = new ChunkedString()
+	>	cstring.append("someString")
+	>	cstring.append("someOtherString")
```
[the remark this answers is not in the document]
Frotty: no it only chunks oversize strings

wurst/network/GamecacheBuffer.wurst
Frotty: yes, the gamecache syncing does not work anymore.

wurst/network/GamecacheKeys.wurst
Frotty: dito

wurst/network/Metadata.wurst
Frotty: Network doesn't work since the patches. This is only used by Network.

wurst/network/SyncSimple.wurst
Frotty: dito

wurst/file/ChunkedString.wurst
```diff
@@ -0,0 +1,86 @@
...
+	>	cstring.flush()
+
+	Read Data Sequentially:
+	>	while cstring.hasChunk()
+	>		let chunk = cstring.readChunk()
+	>		Log.info("read chunk: " + chunk)
```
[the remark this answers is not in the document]
Frotty: iterator could be added

wurst/file/ChunkedString.wurst
```diff
@@ -0,0 +1,86 @@
...
+	which is used to circumvent the maximum string length limit
+	and to prepare data for chunked storage, e.g. for FileIO.
+	Make sure to flush the chunked string when done writing to it,
+	so the unfilled buffer gets written as well.
+
+	Write Data:
+	>	let cstring = new ChunkedString()
+	>	cstring.append("someString")
+	>	cstring.append("someOtherString")
+	>	cstring.flush()
```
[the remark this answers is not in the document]
Frotty: yea, I removed it

wurst/file/ChunkedString.wurst
```diff
@@ -0,0 +1,86 @@
+package ChunkedString
+import Table
```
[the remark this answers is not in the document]
Frotty: string generic overhead very bad

wurst/file/ChunkedString.wurst
```diff
@@ -0,0 +1,86 @@
...
+
+/*	A chunked string splits up big strings into chunks,
+	which is used to circumvent the maximum string length limit
```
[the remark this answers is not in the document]
Frotty: 4099 characters with concatenation, 1024 without.

wurst/file/ChunkedString.wurst
```diff
@@ -0,0 +1,86 @@
...
+	>		let chunk = cstring.readChunk()
+	>		Log.info("read chunk: " + chunk)
+
+	Read Data iteratively:
+	>	for i = 0 to cstring.chunkSize
+	>		let chunk = cstring.getChunk(i)
+	>		Log.info("read chunk: " + chunk)
+*/
+/** Maximum length of one chunk */
+@configurable public constant DEFAULT_CHUNK_SIZE = 200
```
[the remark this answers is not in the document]
Frotty: Needs some testing

wurst/file/FileIO.wurst
Frotty: errror( crashes threads my friend

wurst/file/FileIO.wurst
```diff
@@ -0,0 +1,105 @@
...
+
+	Write a file:
+	>	new File("MyFileName.pld")..write("MyContent")..close()
+*/
+
+/** Abilities which tooltips will be used to load data */
+@configurable constant IO_ABILITIES = [AbilityIds.aerialShackles, AbilityIds.rocketAttack, AbilityIds.callToArms,
+						AbilityIds.calltoArms, AbilityIds.cloudofFog, AbilityIds.controlMagic,
+						AbilityIds.defend, AbilityIds.dispelMagic, AbilityIds.feedbackArcaneTower,
+						AbilityIds.feedback]
```
rhazarian: Hey. It's very cool that you are doing a new FileIO/etc system with the new natives.
I'm using an upgraded FileIO in my map (still with the old syncing) and I have some suggestions.
There are cases when you may want to insert a thousands of lines in a single file - and the game handles this very well, it's tested! :)
You do not need to use a bunch of abilities, you can use a single one and store data into the tooltips of different levels.
This way you could make the max file size easily configurable.

Like this:
```
@configurable public constant DATA_PACKETS_PER_FILE = 64

constant DATA_DUMMY_ID = 'AXDT'
constant DATA_DUMMY_LEVELS = DATA_PACKETS_PER_FILE

@compiletime function generateDummy()
    let dummy = new AbilityDefinitionIllidanChannel(DATA_DUMMY_ID)
        ..setLevels(DATA_DUMMY_LEVELS)

    for i = 1 to DATA_DUMMY_LEVELS
        dummy.setTooltipNormal(i, "")

constant DATA_DUMMY_RAWCODE = compiletime(DATA_DUMMY_ID.toRawCode())

constant DATA_PER_PRELOAD = 259

constant DATA_PADDING_1 = "\")\r\n\tcall BlzSetAbilityTooltip('" + DATA_DUMMY_RAWCODE + "',\""
constant DATA_PADDING_2 = "\",$"
constant DATA_PADDING_3 = ")\r\n//"
constant DATA_FOOTER = "\" )\r\nendfunction\r\nfunction AAA takes nothing returns nothing \r\n//"

public constant MAX_PACKET_LENGTH = compiletime(DATA_PER_PRELOAD - 
                                                DATA_PADDING_1.length() - 
                                                DATA_PADDING_2.length() - 
                                                DATA_PADDING_3.length() -
                                                DATA_DUMMY_RAWCODE.length() -
												5)
```

And then write it like this:
```
Preload(DATA_PADDING_1 + packet + DATA_PADDING_2 + i.toHex() + DATA_PADDING_3)        
```

Where toHex works like this:
```
function int.toHex(int length) returns string
    var temp = this
    var result = ""
    for i = 1 to length
        result = hexs[temp mod 16] + result
        temp = temp div 16
    return result
```

Easily configurable, no limits, efficient.

Frotty: I mostly copied trigger happy's impl, because that works with 1.29 and 1.30 also, not only 1.31 or so. The multi level thing can be done, but it is still limited by the max string limit, so no, not "no limits" and it makes sense to keep it chunked.

rhazarian: Yea. I mean no limits on the total size :)
`i` in the example above was the chunk number.

Frotty: > Yea. I mean no limits on the total size :)

Yes, I ment the total size, which is still limited by the maximum string length.

> A special multilevel ability is also better because you do not need to restore tooltips after reading.

> The suggested method should not also break anything on the previous patches - actually, nothing changes except the ability id and levels.
I can't recall the patch number when I started to use this, but it should be less than 1.31

Before 1.31 you can only read the first ability level, this is why TH did it this way. In 1.29 custom objects can't be read iirc. But I am fine with only supporting 1.31 at this point.

rhazarian: > Before 1.31 you can only read the first ability level, this is why TH did it this way. In 1.29 custom objects can't be read iirc.
Ah, okay, didn't know this, thanks for clarifying.

> Yes, I ment the total size, which is still limited by the maximum string length.
Just wondering - do you mean the total file length? As far as I understand, by keeping the strings chunked the total size of the file should not be limited.

wurst/file/FileIO.wurst
```diff
@@ -13,14 +15,12 @@ import ErrorHandling
...
-/** Abilities which tooltips will be used to load data */
-@configurable constant IO_ABILITIES = [AbilityIds.aerialShackles, AbilityIds.rocketAttack, AbilityIds.callToArms,
-						AbilityIds.calltoArms, AbilityIds.cloudofFog, AbilityIds.controlMagic,
-						AbilityIds.defend, AbilityIds.dispelMagic, AbilityIds.feedbackArcaneTower,
-						AbilityIds.feedback]
+@configurable public constant CHUNKS_PER_FILE = 64
 
 public var CAN_READ_FILES = true
 
+constant DUMMY_ABIL_ID = compiletime(ABIL_ID_GEN.next())
```
rhazarian: This should be a hardcoded id. Otherwise the id may change when you make a new version of the map (adding some abilities), and you won't be able to read the data you wrote in previous versions.

wurst/file/FileIO.wurst
```diff
@@ -0,0 +1,105 @@
...
+		if localPlayer == p
+			PreloadGenClear()
+			PreloadGenStart()
+
+			for i = 0 to buffer.chunkCount
+				if (i >= CHUNKS_PER_FILE)
+					error("FileIO(" + filename + ") ERROR: String exceeds max length (" + (CHUNKS_PER_FILE * buffer.chunkSize).toString() + ").|r")
+
+				let chunk = buffer.getChunk(i)
+				Preload("\" )\ncall BlzSetAbilityTooltip('" + DUMMY_ABIL_ID.toRawCode() + "', \"" + chunk + "\", " + i.toString() + ")\n//")
```
rhazarian: You may want to precompute `toRawCode()`, since writing files (especially big) may be performance sensitive, and `toRawCode` does a loop with divisions inside, and here it's used in a loop, or is the optimizer able to detect it and do it itself, since the `DUMMY_ABIL_ID` is a compiletime value?

rhazarian: The sense in formatting to hex is to keep the length of the chuck number (the level) and the length of the string constant, since the length of the line is limited. So, when it's constant, the exact max size of the chunk can be determined. An integer should be prefixed with `$` for the Wc3 to read it as a hex number. The hex representation is also shorter for big integers, and anyway we have to consider the maximum possible length of the integer when we set the maximum chunk size.

Frotty: I don't think that's necessary

wurst/file/FileIO.wurst
Frotty: good point

wurst/file/FileIO.wurst
Frotty: the optimizer happens after compiletime stuff is over. I don't think toRawCode would bottleneck here, but I added the global.

wurst/file/FileIO.wurst
Frotty: No, I mean the maximum length of  a Jass `string`. With concatenation it's 4k, which usually should be sufficient, but I wanna keep it flexible enough from the start. Also the 4k string can be annoying to handle, because you can't take big substrings and it's generally a hassle, at least that's what I found from my last tests.

rhazarian: Lol, fixed it 😅
As for the length of the strings, I knew this. I was just talking about limits on the file size, which was limited by abilities number.
In my use case I’m writing files with thousands of lines, and the dummy ability has 8k levels 😅

Frotty: Alright, so does anything speak against merging this?

rhazarian: This version looks good to me.
Yes, I'm still using the syncing based on gamecache and units selection, playing on 1.31 - it works.

Frotty: hm okay, I just know that it didn't work anymore in EBR, but this sounds like it's better to deprecate instead of delete after all.

Frotty:
Alright, I will merge this with the deletions, since deprecating the packages would create unavoidable warnings in the stdlib and "updating" the old classes, e.g. `Persistable` isn't easily done because of how the old syncing and data management works.
Most of these packages don't work or only partially work anyway and I don't want people get confused by the similar package names.
The plan is to replace `Persistable` with a new, lightweight `Serializable` class/interface in the future.

PR #305, opened 2019-10-21
make `compiletime` magic function generic

rhazarian:
Would allow to create objects at compiletime with comfort without needing to declare a `compiletime` function with an argument of the needed type each time. What's more, that is the only way to allow to create both an object of type `class A` and an object of type `class B extends A` at compiletime: if you declare `compiletime` functions with arguments of the concrete types, you would not be able to create an object of type B without casting it from A to B, and if you declare generic `compiletime` function, it would be ambiguous with the built-in ones.

### DIFF
--- a/wurst/_wurst/MagicFunctions.wurst
+++ b/wurst/_wurst/MagicFunctions.wurst
@@ -28,23 +28,5 @@ public constant compiletime = false
 /** This is a builtin magic function.
  * It evaluates it's argument at compiletime and replaces the call with the result.
  */
-public function compiletime(int expr) returns int
-	return expr
-
-/** This is a builtin magic function.
- * It evaluates it's argument at compiletime and replaces the call with the result.
- */
-public function compiletime(real expr) returns real
-	return expr
-
-/** This is a builtin magic function.
- * It evaluates it's argument at compiletime and replaces the call with the result.
- */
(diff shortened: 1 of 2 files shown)


Frotty:
Did u test this?

rhazarian:
Yea, I'm using this.

Frotty:
Is this a breaking change in that u need to provide the generic parameter now, or is it backwards compatible?

rhazarian:
Nothing should break if you're using stdlib only - the generic parameter is deduced.

The only cases where it gonna require some extra work is when you have declared a `compiletime` yourself.

I.e. if you previously had:
```
tuple supertuple(int a, int b)

function compiletime(supertuple expr) returns supertuple
    return expr

let a = compiletime(supertuple(2 * 2, 2 + 2))
```

you will be required to delete the `compiletime(supertuple expr)` function, since otherwise it would cause ambiguities for the compiler.

Well, I can't judge, though I think that such cases would be superrare and they don't require a lot of time to fix, the fact that you cannot instantiate a class that extends another class in compiletime (well, you can, if you declare a global, and then make a function with `@compiletime` annotation where you do the assignment, but it's weird) seems to be more important.

Frotty:
yea looks fine, thanks 👍

PR #306, opened 2019-10-22
add fx2 package

Cokemonkey11:


Frotty:
This isn't what I meant. Fx2 should be like Fx, but use an effect only without a unit and offer some neat ways to position the effect, e.g. `lookAt`. Therefore gonna close for now.

PR #309, opened 2019-10-26
FunctionalInterfaces package

rhazarian:
There are many cases when a function accepts a closure. Very often it's just a predicate or a binary function, but you have to declare a new interface each time in each package. These package provides common functional interfaces to use them everywhere instead of declaring new interfaces.

### DIFF
--- a/wurst/_wurst/FunctionalInterfaces.wurst
+++ b/wurst/_wurst/FunctionalInterfaces.wurst
@@ -15,11 +15,6 @@ public interface BiFunction<T:, U:, R:>
 /** Represents an operation upon two operands of the same type, producing a result of the same type as the operands. */
 public interface BinaryOperator<T:> extends BiFunction<T, T, T>
 
-    int a = 0
-
-    construct(int a)
-        print(a)
-
 /** Represents a predicate (boolean-valued function) of two arguments. */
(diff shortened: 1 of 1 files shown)


Frotty:
I'm not sure if this is a good fit for the stdlib.
As I elaborated before, I actually like and support the concept of having specialized functional interfaces.
It improves encapsulation, type safety, stack traces and more, and usually just takes a couple lines and therefore doesn't usually warrant using an external package.
A concrete, wurst-specific issue is that all closures that extend the same functional interface share the same dispatch switch. If you have a hundred closure impls, the dispatch can noticeably affect performance. This problem is much less severe with package specific interfaces.
Also, apart from using the new generics (which are still not done) this just seems like a copy of a small part of https://github.com/theQuazz/wurst-lodash .

rhazarian:
> seems like a copy of a small part of https://github.com/theQuazz/wurst-lodash
Did not know about this package :)
It is a copy of Java's function package actually.

>  all closures that extend the same functional interface share the same dispatch switch. If you have a hundred closure impls, the dispatch can noticeably affect performance.
Oh, didn't think about that. Taking this into consideration, I agree that it's better to stick wih package specific interfaces.

PR #310, opened 2019-10-29
Framehandle package update

GetLocalPlayer:
Spaces replaced with tabs.

- `getFrameByName` was replaced with `getFrame`
- `createFrameByType` was replaced with overloaded `createFrame`
- `setVisibleFor` was replaced with overloaded `setVisible`
- `showFor`, `hideFor` were replaced with overloaded `show` and `hide`
- `name` was replaced with `getName`
- added `enable` `disable` functions
- `minMax` was replaced with `setMinMax`
- added hotdocs to `setLevel`
- `hideOriginFrames(bool)` was repcalsed with `setOriginFramesVisible(bool)`
- added `hideOriginFrames` and `showOriginFrames`
- `boolean` keyword was replaced with `bool`

`_Hadles` package now imports in alphabet order


Frotty:
Some things may depend a bit on taste, like "boolean keyword was replaced with bool", but overall seems useful.
Thanks 👍 
We should probably add some layout helpers that can be used for multiboard and framedefs or so, as I postulated back then for trokkin's multiboard PR.

PR #313, opened 2019-11-01
fix ObjectDefinition in compiletime

rhazarian:
ObjectDefinitions have string keys in the compiler.

The following code does not compile without this fix:

```
let a = compiletime(new UnitDefinition('xxxx', 'hfoo'))
```

The reason is that's the compiler uses string keys, the stdlib uses int keys. It compiles into assigning in string to integer, which fails compilation at pJass stage. This PR fixes the issue.

(Note that it works even without this fix in `@compiletime` functions, because objects created in such functions are not being transmitted to runtime)

### DIFF
--- a/wurst/_handles/Effect.wurst
+++ b/wurst/_handles/Effect.wurst
@@ -20,11 +20,11 @@ public function flashEffect(string path, vec2 pos, real scale)
 public function flashEffect(string path, vec3 pos, real scale)
 	flashEffect(path, pos, scale, GetRandomReal(0, 359).fromDeg())
 
-public function flashEffect(string path, vec2 pos, real scale, angle roll)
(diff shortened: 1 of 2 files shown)


Frotty:
The code snippet contains syntax errors. Also remove the 8b1f7d1 commit please.
I'm not sure what this "fixes" ? if you want to supply "hfoo" as a string, just convert it to an int using `"hfoo".fromRawCode()`.

rhazarian:
Whoops, I have no idea how this commit has got here 🤣 . Gonna remove it a little bit later.

Yea, I see, there is a typo, there should be a closing `'` instead of `"`.

Frotty:
Okay, rebase this branch and I can merge

Frotty:
???

PR #322, opened 2019-12-10
Adding HSV

IgneousRed:


wurst/util/Colors.wurst
```diff
@@ -60,13 +62,77 @@ public tuple color(int red, int green, int blue)
...
+			if g < b
+				h = 1 + (g - b) / delta / 6
+			else
+				h = (g - b) / delta / 6
+		else if g == max
+			h = 1.0 / 3 + (b - r) / delta / 6
+		else
+			h = 2.0 / 3 + (r - g) / delta / 6
+	else
+		h = 0
```
Frotty: Why aren't s and h initialized with these values?

wurst/util/Colors.wurst
```diff
@@ -202,13 +272,20 @@ public function string.toPlayerColor() returns playercolor
 			break
 	return pc
 
+function colorHSV.test()
```
Frotty: should probably be named `assert` instead of test

wurst/util/Colors.wurst
```diff
@@ -219,6 +296,15 @@ init
 	assertTrue(fromHexString("33ccff44") == colorA(204, 255, 68, 51))
 
 	assertTrue(PLAYER_COLOR_RED != null)
+	
+@Test public function testHSV()
+	colorHSV(0, 0, 1).test()
```
Frotty: Can you add an actual use case? Right now I don't see what you would use it for.

wurst/util/Colors.wurst
IgneousRed: If they are, they would be assigned twice if the max and delta are not 0
A small speed optimization, If you are against it, I will change it

wurst/util/Colors.wurst
IgneousRed: Sure

wurst/util/Colors.wurst
IgneousRed: Use what for? colorHSV in general or the tests/asserts?

wurst/util/Colors.wurst
Frotty: Yea, it makes the code uglier and longer, and I highly doubt that it would make a noticeable difference. I also think color transformations usually don't happen in performance critical situations?

wurst/util/Colors.wurst
Frotty: Something that someone would actually use in a map. E.g. have a rgb color, convert it and decrease saturation, then convert back. The test seems very artificial right now.

wurst/util/Colors.wurst
IgneousRed: My opinion is otherwise, but sure, It's your library

wurst/util/Colors.wurst
IgneousRed: Instead of typing tons of manual conversions (as HSVtoRGB and vice versa has alot of edge cases) this checks both and is alot shorter. Proving that both are working as intended. I don't see the issue as this is the most efficient way while using 50 lines less

wurst/util/Colors.wurst
Frotty: If you can provide a real world example where this matters (i.e. noticeable improvement), feel free to do so.
You found out yourself that we value readability and safety over these microop optimizations shenanigans.

wurst/util/Colors.wurst
Frotty: Yea, but I'm not talking about asserting correctness.

PR #325, opened 2019-12-21
  fix HashList not resize

yefq:
before: method ".set" always not resize list.
fixed: resize list when use a index out for this list.

this fix make "discrete list" run better.


Frotty:
Hint: To avoid all these commits you should start a fresh branch for each PR.

I guess this can be useful, but the usage is pretty unobvious. The hot doc states:

> Set an existing element at the given position to a new value

So this would not be intended behavior currently.
I think you should change the hot doc and add a unit test 🌵 .

yefq:
I will resubmit a PR tomorrow, thank you for your advice.
I'm going to get some sleep (because it's already 3:30 in China)...

yefq:
resubmit later.

PR #326, opened 2019-12-22
Base64 package

rhazarian:
Base64 package is intended to be used for serialization/deserialization of arbitrary data. It's highly optimized (as much as I could while keeping it's nice to use - the implementation might be ugly though), because I believe that it could be a bottleneck - especially the loops processing bytes/string chunks when data is quite big, e.g. 500 KB - it's pretty much for Wc3, and I have such cases (e.g. serializing the state of the whole map as a save).
I've also fixed some memory leaks & encapsulation in ChunkedString & FileIO packages.


Frotty:
Hi, thanks for a contribution, a few things:

> It's highly optimized (as much as I could while keeping it's nice to use - the implementation might be ugly though), because I believe that it could be a bottleneck - especially the loops processing bytes/string chunks when data is quite big, e.g. 500 KB - it's pretty much for Wc3, and I have such cases (e.g. serializing the state of the whole map as a save).

I don't see that as a valid reason to make something ugly optimized, since save/load stuff isn't executed periodically nor performance critical. imho the better approach is to defer the calculations and work with closure callbacks.

> I've also fixed some memory leaks & encapsulation in ChunkedString & FileIO packages.

Those looks good, but should be a separate PR and merged before, as it bloats this PR unnecessarily.

Regarding the general idea of using Base64 - does it really make sense compared to just using [arithmetic encoding](https://en.wikipedia.org/wiki/Arithmetic_coding)? I am using a port of pipedream's original script for my save data https://github.com/Frotty/Frentity/blob/master/wurst/util/Encoder.wurst
I haven't gotten around to making proper stdlib packages.

rhazarian:
> Those looks good, but should be a separate PR and merged before, as it bloats this PR unnecessarily.

Sure.

> Regarding the general idea of using Base64 - does it really make sense compared to just using arithmetic encoding? I am using a port of pipedream's original script for my save data https://github.com/Frotty/Frentity/blob/master/wurst/util/Encoder.wurst
I haven't gotten around to making proper stdlib packages.

I think these are a little bit about different things. Base64 is about writing bits as a string and decode them back, and Arithmetic Encoding is about compressing bits into bits.

As I have understood, I're using the fact that your ints are not bigger than ~60, so that you can make a bijection between each int and a char, while the Base64 encoder supports completely arbitrary data.
Correct me if I'm wrong.

> I don't see that as a valid reason to make something ugly optimized, since save/load stuff isn't executed periodically nor performance critical. imho the better approach is to defer the calculations and work with closure callbacks.

I had a look at your SeqWorker, it may be an alternative and I've thought about this, possibly making this as an alternative to the blocking implementation a little bit later.

I agree that optimizations may be not worth when it's 10-20% only, but talking about, for example, the execution loop when decoding bytes, the actual code for decoding takes like 5 lines of compiled code, while with using the stdlb executeWhile with lambdas it would took X times more (X > 2), because the dispatch for one lambda may take more than 5 lines already, and we have 2 there. It seems to me that we should not make players wait for 30 secs instead of 10 secs just to make the code here a little bit more nice.

If you see any useless optimization, please, point them out.

rhazarian:
Okay, I've rebased on the fresh master.

Frotty:
Looking good, can you add a usage example for encoder/decoder at the top?

rhazarian:
Sure, will do.

rhazarian:
I've added some brief examples.

wurst/file/ByteBuffer.wurst
```diff
@@ -0,0 +1,145 @@
...
+
+import Table
+import Bitwise
+
+/*	ByteBuffer is a utility class that allows you to write/read bytes, shorts and integer to/from it.
+	It may be useful for compacting bytes into integers or decomposing integers into bytes.
+*/
+
+public class ByteBuffer
+	private var storage = new Table()
```
Frotty: why are you using a Table if you only load/store ints?

wurst/file/Base64.wurst
```diff
@@ -0,0 +1,284 @@
...
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+]
+
+public class Base64Encoder
+    private ChunkedString chunkedString
+
+    // We append only one char each time, and strings in WC3 get fully copied on each concatenation.
+    // To optimize it and lower the times of copying quite big chunks, append by smaller chunks.
+    private var stringBuffer = ""
```
Frotty: Isn't this a bit superflous to the buffer inside the ChunkedString?

wurst/file/Base64.wurst
```diff
@@ -0,0 +1,284 @@
...
+    private var chars = 0
+    private static constant MAX_CHARS = 32
+
+    private var buffer = 0
+    private var bytes = 0
+    private static constant MAX_BYTES = 3
+
+    construct()
+        this.chunkedString = new ChunkedString()
```
Frotty: initialize the member directly

wurst/file/Base64.wurst
```diff
@@ -0,0 +1,284 @@
...
+
+    /** This constructor consumes the ByteBuffer. If you want to save it, use the 'write' method. */
+    construct(ByteBuffer data, int maxChunkLength)
+        this.chunkedString = new ChunkedString(maxChunkLength)
+
+        writeData(data)
+        destroy data
+
+    ondestroy
+        if chunkedString != null
```
Frotty: if you always initialize the member anyway, it will be constant and you can get rid of this nullcheck

Frotty:
> As I have understood, you're using the fact that your ints are not bigger than ~60, so that you can make a bijection between each int and a char, while the Base64 encoder supports completely arbitrary data.

It depends how it is used. You can use it very similar to Base64 and simply pass known maximum bounds of any saved value, which then allows for saving pretty much arbitrary data, but also using the optimal, least amount of "character space" in the output string. E.g. if you know certain numbers will always be small, you can encode both of them in a bigger number. Or another examples are shorts in your code, which still use full int space since they are just padded.

My examples are not directly comparable because I first serialize into a json-like format, then LZW compress it, and then encode the compressed data arithmetically, where the maximum of each encoded value is the amount of different characters in the LZW output.

> I had a look at your SeqWorker, it may be an alternative and I've thought about this, possibly making this as an alternative to the blocking implementation a little bit later.

Please check the creation dates on that stuff, it's mostly old and legacy - I wouldn't use it as example and would myself implement many things differently by now.

I do prefer this over the existing encoder which just puts data into serial blocks `"0.12|sadasd|123"` and would like to deprecate more of the old network stuff so we just have one good solution, but I'm not sure if it is the best one.

wurst/file/Base64.wurst
rhazarian: This nullcheck is here because the string is returned to the user in 'intoData' method, so there it's set to null and prevent it from being destroyed.

wurst/file/Base64.wurst
rhazarian: The thing I'm worried about is the following:
Imagine we have the buffer inside the ChunkedString with 160 chars. Here we know we append by exactly one symbol. We append a char 32 times. So the buffer inside the string becomes 161 chars, 162 chars, 163 chars...192 chars long... And copies each time entirely. So 161 chars is copied in the first append, 162 in the second, 163 in the third...192 chars in the end.
If we append to this buffer previously:
1 char get copied, 2 chars get copied...32 chars get copied... This buffer is full. Append to the ChunkedString buffer: 160 + 32 chars get copied one time.
Look at the difference here: 1+2+...32 + 32 + 160 = 720 character copies with this optimization VS 160+161+...200 = 5648 character copies without.

wurst/file/ByteBuffer.wurst
rhazarian: Could have used HashMap, but (I'm a little bit paranoid probably :D) just chose the simplest thing satisfying the needs (since I don't ever remove the elements, don't need to increment/decrement the size, etc.) + tried to avoid some extra overhead the the HashMap offers, i.e. size tracking with `has` checks on the every `put` method. Should not matter a lot, though may when there is over 500k write/read ops (for 500 Kb file for example).

wurst/file/Base64.wurst
rhazarian: Hmm, could do it that way.

Don't you think though that it may provide some possibilities for the class user than actually should not be provided?
E.g. if he/she changes to chunksize somewhere in the middle of writing, while the class will continue to working correct, the result may be a little bit strange + you won't be able to be sure that chunkSize is the same for all chunks, which is used in the length computing, which is used even in this package.

Also the constructors with ByteBuffers are not actually needed - could remove them, because the extension functions that write everything from a ByteBuffer already exist. Though they don't consume it.
We can ask the the user to destroy the Buffer themselves, or make an additional extension function which consumes the provided buffer. Or just go with these constructor. What do you think is the better approach? 😃

rhazarian:
Thanks for the explanation.

> Or another examples are shorts in your code, which still use full int space since they are just padded.

They are actually compacted into an integer thanks to the ByteBuffer functionality.
`writeByte(a)..writeShort(b)..writeByte(c)..readInt()` would return an integer which the byte, the short and the byte.

> but I'm not sure if this is the best one yet.

Ain't sure aswell 😆 
Well, as far as I've looked over the net, if our aim is to serialize any data in ASCII text (since it's the only way we can put it into the Preload functions), one of the best approach is to compress the bytes first, e.g. with LZW, and then Base64-encode the output bytes.

wurst/file/Base64.wurst
Frotty: > Probably a consuming extension function may be good, just need to name in a way that there will be no confusion.

Yeah.

Frotty:
Please make the remaining changes as discussed.

> Well, as far as I've looked over the net, if our aim is to serialize any data in ASCII text (since it's the only way we can put it into the Preload functions)

Yes probably. I do plan on re-adding a `Serializable` interface for general class de/serialization which will probably use a json-like format.

> one of the best approach is to compress the bytes first, e.g. with LZW (or arithmetic encoding), and then Base64-encode the output bytes.

I mostly chose LZW for ease of implementation and because the bit natives haven't been added yet.
Now we can probably implement something more elegant.
In any case I suppose base64 is useful since it's a known format.
I hit many limits with my save stuff - perhaps I can try this in EBR after merge.

rhazarian:
> Please make the remaining changes as discussed.

Moved the logic of the constructors into extension functions.

wurst/file/Base64.wurst
```diff
@@ -0,0 +1,270 @@
+package Base64
+
+import Bitwise
+import StringUtils
+import ByteBuffer
+import ChunkedString
+import Execute
+import ErrorHandling
+
+/*  Base64 package provides an encoder/decoder pair that allows serializing and deserializing of any data.
```
Cokemonkey11: Phrasing: encoder/decoder pair sounds like the library provides function to emit a linked encoder-decoder pair like [channel](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html). Better: "provides encoding and decoding function for the common base64 format"

Phrasing: "serializing and deserializing of any data" - what does *any data* mean? Wurst is a strongly typed language: can you rephrase this to explain the type relationship?

wurst/file/Base64.wurst
```diff
@@ -0,0 +1,270 @@
...
+
+    Encode Data iteratively:
```
Cokemonkey11: Phrasing: I think `iteratively` makes this less clear since that's a bit of an overloaded term. I'd describe this as providing a builder pattern.

wurst/file/Base64.wurst
```diff
@@ -0,0 +1,270 @@
...
+    > encoder.writeInt(100500)
+    > let data = encoder.intoData() // the encoder is consumed, no need to destroy it
+
+    Decode Data iteratively:
+    > let decoder = new Base64Decoder()
+    > while data.hasChunk()
+    >   decoder.append(data.readChunk())
+    > let bytes = decoder.intoData() // the decoder is consumed, no need to destroy it
+    > let byte = bytes.readByte()
+    > let number = bytes.readInt()
```
Cokemonkey11: This API block is a bit bizarre IMO. Is `decoder` a mutable, stateful iterator for `data`?

wurst/file/Base64.wurst
```diff
@@ -0,0 +1,270 @@
...
+
+    This package also provides convenient methods for ChunkedStrings and ByteBuffers.
+
+    Encode Data:
+    > let data = bytes.encodeBase64()
```
Cokemonkey11: Types?

wurst/file/Base64.wurst
```diff
@@ -0,0 +1,270 @@
...
+
+    Decode Data:
+    > let bytes = data.decodeBase64()
```
Cokemonkey11: Types?

wurst/file/Base64.wurst
```diff
@@ -0,0 +1,270 @@
...
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+]
+
+public class Base64Encoder
+    private ChunkedString chunkedString
+
+    // We append only one char each time, and strings in WC3 get fully copied on each concatenation.
```
Cokemonkey11: This comment is useful but doesn't describe the function of `stringBuffer`

wurst/file/Base64.wurst
```diff
@@ -0,0 +1,270 @@
...
+        this.chunkedString = new ChunkedString()
+
+    construct(int maxChunkLength)
+        this.chunkedString = new ChunkedString(maxChunkLength)
+
+    ondestroy
+        if chunkedString != null // The string is null after calling `intoData()`.
+            destroy chunkedString
+
+    // It's way too much overhead in dispatching two lambdas per each byte, 'inlining' them to deal with this.
```
Cokemonkey11: > It's way too much overhead

Justification?

> 'inlining' them

Who is inlining them, where?

wurst/file/Base64.wurst
```diff
@@ -0,0 +1,270 @@
...
+    private function encode(int byte, int count)
+        var remaining = byte
+
+        for i = 0 to count
+            let c = remaining.bitAnd(compiletime("11111100 00000000 00000000".fromBitString())).shiftr(18)
+            append(CHARMAP[c])
+            remaining = remaining.shiftl(6)
+
+    /** Writes an unsigned byte to be serialized into Base64. 
+    You must be sure that the provided integer is in the range [0, 255]. */
```
Cokemonkey11: > You must be sure that the provided integer

Ideally find a way to add checks that can be optimised away

wurst/file/ByteBuffer.wurst
```diff
@@ -0,0 +1,145 @@
+package ByteBuffer
+
+import Table
+import Bitwise
+
+/*	ByteBuffer is a utility class that allows you to write/read bytes, shorts and integer to/from it.
```
Cokemonkey11: What is the analog class in - say - Java?

What's the endianness?

wurst/file/ByteBuffer.wurst
```diff
@@ -0,0 +1,145 @@
...
+	It may be useful for compacting bytes into integers or decomposing integers into bytes.
```
Cokemonkey11: Is there a use case for writing bytes and reading bytes, or is this a type conversion utility?

wurst/file/ByteBuffer.wurst
Cokemonkey11: Use HashMap

Frotty:
@rhazarian will you update/respond?

rhazarian:
Sorry too busy rn, need 1 more week or so

rhazarian:
Today's Reforged update has finally fixed the tooltips so this thing should be usable when merged, I'll adopt & test this for my map now, then update it here.

Frotty:
great :+1:

wurst/file/Base64.wurst
rhazarian: You just have to know what does it compile to. A big switch with elseifs per each possible instance and a TriggerExecute. It's obvious that we do not want to do 10 ifs and a TriggerExecute call per each byte.
That was meant to describe why I'm not using stdlib's executeWhile here and reimplement the functionality, by 'inlining' the lambdas into the executeWhile. I'll make the description better.

wurst/file/Base64.wurst
rhazarian: Add a comment that these chunks are actually stored in the buffer, but it was obvious IMO.

wurst/file/ByteBuffer.wurst
rhazarian: There is. Base64Encoder/Decoder use it to write bytes there and read bytes from there to give these bytes to the user then.

wurst/file/Base64.wurst
rhazarian: Attempted to make it better.

wurst/file/Base64.wurst
rhazarian: Didn't want to make a bad example by now using `let` so added types in comments

wurst/file/Base64.wurst
rhazarian: No, the decoder only accepts strings to be decoded and returns a `ByteBuffer` with data.

wurst/file/ByteBuffer.wurst
rhazarian: `ByteBuffer` is something similar in Java, though has more functionality.
Integers bigger than byte are treated as in BigEndian.

wurst/file/Base64.wurst
rhazarian: They're private here. ByteBuffer have similar public functions and I've added checks there using SafetyChecks package.

rhazarian:
Tested it in this map in save/load system.

Also I'm not sure if empty tooltips are safe enough so changed it to a one space.

wurst/file/Base64.wurst
Cokemonkey11: The first change looks good, but what does `bytes` mean? Jass doesn't have a `byte` type

wurst/file/Base64.wurst
Cokemonkey11: IMO you can just delete this comment

wurst/file/ByteBuffer.wurst
Cokemonkey11: I think the documentation needs that context, not the PR

Cokemonkey11:
How will the program behave now if it encounters an empty string?

LGTM

✨ 🚀 
🔥 ✨

rhazarian:
It could only encounter an empty string if the Preload file contained an empty string, which should be impossible. Even if it was there, then the empty string would be just appended to the rest so no effect /shrug. This place should be safe to change

Frotty:
Please check unresolved conversations and answer/resolve accordingly 🚀

rhazarian:
Further improved docs according to Cokemonkey's suggestions.

Frotty:
Sorry kinda forgot to merge this 😅 
Thanks a lot 🍰

PR #328, opened 2019-12-22
update HashList

yefq:
set an element at an index out of size.
auto resize this list, and insert elements use null.


Frotty:
:+1:

PR #330, opened 2019-12-22
Prettify ChunkedString and FileIO

rhazarian:
Encapsulation, memory leaks and useless actions fixed.

### DIFF
--- a/wurst/file/SaveLoadData.wurst
+++ b/wurst/file/SaveLoadData.wurst
@@ -14,7 +14,7 @@ import ErrorHandling
 	>	myPlayer.loadData("myFileName") (status, data) ->
 	>		switch status
 	>			case SUCCESS
-	>				Log.info("Load successful. Chunks: " + data.chunkCount)
+	>				Log.info("Load successful. Chunks: " + data.getChunkCount())
 	>			default
 	>				Log.error("Error occured: " + status.toString())
 */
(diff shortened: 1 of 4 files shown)


wurst/network/SyncSimple.wurst
```diff
@@ -103,4 +104,3 @@ init
 				syncQueue.dequeue()
 				destroy syncData
 		checkQueue()
-
```
Frotty: dont remove EOF newlines. editorconfig should enfore this if u use vscode.

wurst/file/SaveLoadData.wurst
```diff
@@ -46,7 +46,7 @@ public interface LoadListener
 public function player.saveData(string slotName, string data)
 	let buffer = new ChunkedString(DEFAULT_CHUNK_SIZE)
 	buffer.append(data)
-	this.saveData(slotName.endsWith(".pld") ? slotName : slotName + ".pld", buffer)
```
Frotty: Last time i checked, `.pld` ending was required for the fileIO to work.
Have you tested this?

rhazarian: It is required, but it’s handled by the other overload of the function - the one taking a ChunkedString. The edited overloaded basically just converts a string into a ChunkedString and the file ending was handled twice.

rhazarian: It is required, but it’s handled by the other overload of the function - the one taking a ChunkedString. The edited overloaded basically just converts a string into a ChunkedString and the file ending was handled twice.

Frotty:
Looks good, but I'm not a big fan of the breaking getter changes.

rhazarian:
I can remove it if you want.
Though I think that it may be a good moment to change it now, because I guess none have used it a lot (if used at all) yet, and now the implementation is exposed and it blocks possible future changes

Frotty:
Thanks 👍 
Make sure to update your other PR please.

PR #331, opened 2019-12-25
Add string.isLower() & string.firstLower()

rhazarian:
There are such functions for uppercase, no idea why there was no lowercase counterparts.


Frotty:
I suppose they don't exist, because you can simply negate the `upper` variants.
The `lower` ones could also be implemented that way.

rhazarian:
Hmm, that may be true talking about ‘isLower’.
You can’t negate ‘firstLower’ though since it modifies the string :)

Frotty:
Ah true. That one seems good then.

PR #333, opened 2019-12-25
Add executeUnsafe to Execute package

atmno:
A handy wrapper to access executeForce.forEach.

execute is an overkill for some simple tasks such as static function, map init..
Using execute in multiple places will generate closures and caused some overhead.

### DIFF
--- a/wurst/closures/Execute.wurst
+++ b/wurst/closures/Execute.wurst
@@ -96,7 +96,7 @@ public function try(ForForceCallback c) returns boolean
 	pushCallback(c)
 	let suppressErrors = suppressErrorMessages
 	suppressErrorMessages = true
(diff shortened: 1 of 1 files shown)


Frotty:
Hi, thanks for your PR. However I'm not sure if that makes sense.

> execute is an overkill for some simple tasks such as static function, map init..
> Using execute in multiple places will generate closures and caused some overhead.

Can you give me a real world example where this matters?
We value safety over a few saved lines of generated code. This would break as soon as there is a nested execute call, which you might not expect.
If you want to do unsafe things, you should probably just use your own Force.

atmno:
I am writing some performance critical task..
Just use execute is a better idea, I think too much, close

PR #335, opened 2019-12-29
Add .sync method for primitives

rhazarian:
Async bools/ints/reals can come from custom frames, this introduces handy wrappers of string.sync for these.


Frotty:
I suppose one is expected to wrap sync data in a serialized class - but sure, this makes sense.

Frotty:
Thanks 👍

PR #340, opened 2020-01-28
More unit extension functions

GetLocalPlayer:
addAbility(id, level)
addPermanentAbility(id)
addPermanentAbility(id, level)


Frotty:
Thanks 👍

PR #342, opened 2020-02-03
Added setter for the maxStack item field

Jaccouille:
Reforged added a field to the item object definition allowing us to set the number of stack for an item


Frotty:
Thanks. I suppose this can be merged without new natives and simply won't have an effect. :+1:

PR #343, opened 2020-02-12
Adapt StdLib to new typeclass features

peq:
Trying out how rewrite of StdLib with new generics and type classes would look like (see https://github.com/wurstscript/WurstScript/pull/931).

Feedback and tests with users of the library are very welcome.

### DIFF
--- a/wurst/Wurst.wurst
+++ b/wurst/Wurst.wurst
@@ -9,3 +9,4 @@ import public Wurstunit
 import public TypeCasting
 import public Colors
 import public Annotations
+import public BasicTypeClasses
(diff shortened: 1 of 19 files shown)


Frotty:
As mentioned I'm not too big of a fan of `public implements` as it sounds wrong.
`implementation` or `trait` are probably more suited.

rhazarian:
Probably it's performance-wise better to introduce a typeclass with methods to write and read the type to/from the native hashtable, so it would be possible to use HashMap with native types and strings without any overhead, as they would use their own methods, like saveString, loadString, etc.

peq:
Yes, that's a good idea. Just have to figure out how to do this so that it's not necessary to implement two type classes for every type that does not have a hasmap native.

PR #344, opened 2020-02-17
Make some constants public

rhazarian:
Someone may want to refer to these constants to protect the dummy units from some sort of their own unit indexer/processor or to preload the abilities when there are a lot of levels.


Frotty:
@rhazarian when resolving conflicts just now I see that u changed the set tooltip from `""` to `" "` at some point? What exactly is intended/correct?

rhazarian:
`" "` is correct. Empty strings are tricky, at least in one patches that I tested it seemed that the tooltip doesn't change at all when trying to make it empty, so that when you read a longer file after a shorter one, you also read a portion of the previous file data at the end. One space should be safer.

Frotty:
But does the `FAIL_FILE_EMPTY` mode work if there is a 1length blank string in the chunkedstring buffer?

rhazarian:
The space won’t get to the buffer, as it is treated as the end of file

Frotty:
Alright thanks 👍

PR #345, opened 2020-02-21
Hotfix ChunkedString

rhazarian:
readIndex is -1 by default, and chunkCount is 0 by default, so an empty ChunkedString returns true on hasChunk.

### DIFF
--- a/wurst/file/ChunkedString.wurst
+++ b/wurst/file/ChunkedString.wurst
@@ -52,7 +52,7 @@ public class ChunkedString
 			chunkCount++
 
 	function hasChunk() returns boolean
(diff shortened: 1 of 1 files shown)


rhazarian:
If it's preferable to keep some kind of compatibility/have an empty chunk at the end of the string, another solution would be to use .length() method when determining whether the file exists in SaveLoadData package. Right now it never returns FAIL_FILE_EMPTY, it succeeds on empty files.

rhazarian:
Oh yes, I guess it would be better to keep it so that a ChunkedString always has at least one chunk. I'll fix the problem in another PR, I'm sorry.

PR #346, opened 2020-02-24
Fix reading empty files

rhazarian:
Even an empty ChunkedString has one (empty) chunk, so you should check it's length instead of checking whether it has a chunk.


Frotty:
iirc i tested this, i suppose this changed due to your changes to ChunkedString. Seems to be the better option anyway though.
Thanks 👍

PR #348, opened 2020-03-06
Add option to autopreset normal tooltips

rhazarian:
When making a lot of custom spells, especially based on standard ones, it's very annoying to set the normal tooltip similar to standard one by hands for each ability. It's also annoying to register a tooltip generator, start and stop listening just for normal tooltip... Since the extended one is done by hands if you want to follow the standard format. This adds an option to just make normal tooltips for all spells which look exactly like the standard one automatically, without having need to call any special methods.

### DIFF
--- a/wurst/objediting/AbilityObjEditing.wurst
+++ b/wurst/objediting/AbilityObjEditing.wurst
@@ -14,6 +14,8 @@ import Annotations
 	property names to have spaces. */
 @configurable constant USE_PROPERTY_SPACING = true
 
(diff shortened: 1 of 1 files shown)


Frotty:
I guess it makes sense to have something like this, but I think I am not too big of a fan of this implementation.

rhazarian:
What would you change/improve?

Krypt0n:
does not seem to be content for a stdlib to me

Cokemonkey11:
@Frotty I recommend closing this

Frotty:
As others have suggested, this doesn't seem like a very good fit. Closing.

PR #349, opened 2020-03-07
Don't require lvl in buffs editing

rhazarian:
In fact, buffs' parameters do not depend on levels. You cannot specify levels of a buff and it only uses it's level 1's parameters, so it's stupid to have to specify the level every time. This PR provides additional convenience overloads.


Frotty:
> You cannot specify levels of a buff and it only uses it's level 1's parameters

Are you sure?
Otherwise looks good 👍

rhazarian:
> Are you sure?

Yea, I'm only setting 1'st level's fields and it works for my spells on any level.
And here it's like it looks in the World Editor:
https://imgur.com/OtZUiBA
Nothing about the levels.

Frotty:
Okay thanks 👍

PR #350, opened 2020-03-12
HOTFIX reading from ByteBuffer

rhazarian:
I'm sorry for the bug I've missed when submitted the ByteBuffer :) Here is the fix. Without it, reading from it in a while loop with `hasByte()` condition works wrong.


rhazarian:
Uh, sorry, needs some further debugging.

PR #351, opened 2020-03-12
[HOTFIX] Reading from a ByteBuffer with a while loop

rhazarian:
I'm sorry for the bug I've missed when submitted the ByteBuffer. :S Here is the fix. Without it, reading from it in a while loop with hasByte() condition works wrong: you could miss some bytes.


Frotty:
Then add a unit test please to catch that problem.

rhazarian:
Good point, here is the test.

Frotty:
Thanks 👍

PR #353, opened 2020-05-23
AbilityDefinitionBeserk typo

GetLocalPlayer:


Frotty:
The typo is from the original game's slk files and has been mentioned a few times.

GetLocalPlayer:
Doesn't  that change the fact that users will be looking for `Berserk` definition and not `Beserk`? Or do you mean these generated definitions must not be changed manually?

Frotty:
Just mentioning, I suppose this "duplicate" can be useful, thanks.

PR #355, opened 2020-06-17
added error print if trying to address a frame by name with exceeding…

Krypt0n:
… maximum name length allowed by wc3:api


Frotty:
Thanks 👍

PR #360, opened 2020-12-20
Added unit test for `length` properties of constant arrays.

jonathanmclaus:


jonathanmclaus:
Isn't it bad practice to leave the print in? I was going to remove that.

PR #363, opened 2020-12-26
Updated Trollbatrider soundset

Jaccouille:
Look like blizzard changed the string used to reference Troll batrider soundset.


Frotty:
Do you have us english version?

Jaccouille:
I changed text & spoken language from French to English through the battlenet launcher like a year ago if that answer your question.

Frotty:
I meant cuz of the comment "us only", dunno what exactly that means tho.

PR #365, opened 2021-01-05
Removed superfluous variable in LinkedList.joinBy.

jonathanmclaus:
This variable isn't needed. It's always `true` until after the if-condition is evaluated in the final step of the for-loop.


Frotty:
Is there a test for this?

jonathanmclaus:
@Frotty there is already a test for this [here](https://github.com/wurstscript/WurstStdlib2/blob/2134995ba3b9904603ef8e916853d45941e50aff/wurst/data/LinkedListTests.wurst#L219) and it still passes.

PR #367, opened 2021-01-11
UnitObjEditing public import in HeroPreset

Overkane:
If you are using HeroPreset.wurst, you can't use stuff like TargetsAllowed or ArmorType and etc. which are publicly imported in UnitObjEditing.wurst. You have to import UnitObjEditing.wurst again or import TargetsAllowed.wurst and ObjEditingCommons.wurst separately.
This fix solves the problem.


Frotty:
👍

PR #369, opened 2021-01-30
Changed texttag shortcuts to output the tag.

jonathanmclaus:
It would be nice for these to output the text tag in case further modification is desired, e.g. filtering visibility.


Frotty:
👍

PR #370, opened 2021-01-30
Added utility methods to texttag.

jonathanmclaus:


Frotty:
I think "attach" is a little misleading because the textag does not stay attached to the unit?
Hotdoc would be appreciated.

jonathanmclaus:
Do you have any suggestions for the name? I could just switch to `.for` with a hotdoc.

PR #372, opened 2021-02-01
Added some boolean operations.

jonathanmclaus:
XOR doesn't have an operator and I figure this is the next best thing.


Frotty:
Could be useful, thanks 👍

PR #374, opened 2021-02-08
Added BuffIds class

Jaccouille:
Simply dumped the name & ids of object editor buffs from 1.32 patch.


Frotty:
Nice, can u fix build errors? 3 variables are named "unknown"

Frotty:
Great 👍

PR #377, opened 2021-02-27
[pre1.29] fix LinkedList.removeAt

isaced:
This problem can cause compilation failure, keep the same as the master branch.

https://github.com/wurstscript/WurstStdlib2/blob/master/wurst/data/LinkedList.wurst#L127


Frotty:
👍

PR #381, opened 2021-03-24
expanded OKSKEY_META

Krypt0n:


Frotty:
nice thanks

Krypt0n:
just something, thought i would make sense here, as i played around with metas and hotkeys anyways

PR #384, opened 2021-04-04
Update HeroPreset.wurst

Overkane:
If you wanna add some normal abilities to hero, like lightning attack or just add some starting abilities, then you have to reset normal abilities after hero builds since inventory normal ability hard coded in hero build func.
With this change you just can add normal abilities similar to hero abilities.


Frotty:
👍

PR #386, opened 2021-04-05
introduce commaList(LinkedList<int>) and commaList(LinkedList<string>)

### LINKED ISSUE #382: commaList support LinkedList arg
```
public function commaList(vararg int ids) returns string
	let list = new LinkedList<int>
	for id in ids
		list.add(id)
	let result = list.joinBy((int i) -> i.toRawCode(), ",")
	destroy list
	return result

public function commaList(vararg string values) returns string
	let list = new LinkedList<string>
	for v in values
		list.add(v)
	let result = list.joinBy(",")
	destroy list
	return result
```

Should refactor this a bit to commaList(LinkedList<string>) <- commaList(LinkedList<int>) <- commaList(vararg int ids)

Cokemonkey11:
Closes #382

### DIFF
--- a/wurst.build
+++ b/wurst.build
@@ -1,2 +1,3 @@
 ---
 projectName: Wurst Standard Library 2
+dependencies: []
--- a/wurst/dummy/DummyRecycler.wurst
+++ b/wurst/dummy/DummyRecycler.wurst
@@ -204,9 +204,12 @@ init
 	..setAttack1DamageSidesperDie(1)
 	..setAttack1ProjectileSpeed(INT_MAX)
 	..setAttack1Range(INT_MAX)
-	..setAttack1TargetsAllowed(commaList(TargetsAllowed.air, TargetsAllowed.debris, TargetsAllowed.ground, TargetsAllowed.invulnerable, TargetsAllowed.item_t, TargetsAllowed.structure, TargetsAllowed.vulnerable, TargetsAllowed.ward))
+	..setAttack1TargetsAllowed(
+		commaList(TargetsAllowed.air, TargetsAllowed.debris, TargetsAllowed.ground, TargetsAllowed.invulnerable,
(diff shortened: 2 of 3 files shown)


wurst.build
```diff
@@ -1,2 +1,3 @@
 ---
 projectName: Wurst Standard Library 2
+dependencies: []
```
Cokemonkey11, line 3: Drive-by fix: without this line, `grill install` creates a copy of stdlib in `_build`

wurst/dummy/DummyRecycler.wurst
```diff
@@ -204,9 +204,12 @@ init
 	..setAttack1DamageSidesperDie(1)
 	..setAttack1ProjectileSpeed(INT_MAX)
 	..setAttack1Range(INT_MAX)
-	..setAttack1TargetsAllowed(commaList(TargetsAllowed.air, TargetsAllowed.debris, TargetsAllowed.ground, TargetsAllowed.invulnerable, TargetsAllowed.item_t, TargetsAllowed.structure, TargetsAllowed.vulnerable, TargetsAllowed.ward))
+	..setAttack1TargetsAllowed(
+		commaList(TargetsAllowed.air, TargetsAllowed.debris, TargetsAllowed.ground, TargetsAllowed.invulnerable,
```
Cokemonkey11, line 208: Drive-by fix: minor tidying to use 120-char line limit

Happy to revert if needed

wurst/objediting/ObjectIds.wurst
```diff
@@ -49,21 +49,29 @@ public function string.fromRawCode() returns int
 public function int.toRawCode() returns string
 	return toRawCode(this)
 
+/** Consumes the provided list, producing a string joined by ','. */
```
Cokemonkey11, line 52: "Consumes" is used elsewhere in codebase, I think it's clear enough

PR #387, opened 2021-04-05
[knockback3] knockback3 ignore air units by default

### LINKED ISSUE #385: Knockback3 doesn't behave appropriately with flying units

Cokemonkey11:
Closes #385

The underlying issue here is that SetUnitFlyHeight always obeys unit min flyheight.

Note that Get/Set min fly height is not working with new blizzard natives, so there's no safe way to knockback3 flying units.

Another reasonable solution here would be to automatically provide a 2D knockback for flying units

### DIFF
--- a/wurst/util/Knockback3.wurst
+++ b/wurst/util/Knockback3.wurst
@@ -36,6 +36,10 @@ import Objects
 public interface TerrainZProvider
 	function get(vec3 where) returns real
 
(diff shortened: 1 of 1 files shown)


Frotty:
`   constant unittype UNIT_TYPE_FLYING                      = ConvertUnitType(3)`

PR #391, opened 2021-04-11
fix DelayNode

Jampi0n:
The queue of DelayNodes is not setup correctly. This results in a wrong order of DelayNodes and next references being null for intermediate nodes.

DelayNode makes use of timers, so unit tests are not possible. For showing correctness and showing the errors of the previous implementation I made a modified version with a unit test here: https://bin.wurstlang.org/jajiwenuwe.wurst
This makes it easy to see the issues of the old implementation without the need to run a map.

The unit test first uses the new implementation and afterwards the old one. The old implementation will result in a nullpointer exception and wrongly sorted queue.

### DIFF
--- a/wurst/dummy/DummyRecycler.wurst
+++ b/wurst/dummy/DummyRecycler.wurst
@@ -102,30 +102,22 @@ class DelayNode
 			DelayNode tmp = first
 			while tmp.next != null and tmp.delayTime < this.delayTime
 				tmp = tmp.next
-			if tmp.delayTime < this.delayTime
-				if tmp == last
-					tmp.next = this
-					this.prev = tmp
-					last = this
-				else
-					tmp.next.prev = this
-					this.next = tmp.next
-					tmp.next = this
-					this.prev = tmp
-			else
+			
+			if tmp.delayTime < this.delayTime // this.delayTime has latest delayTime -> place last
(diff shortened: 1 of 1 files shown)


Jampi0n:
I got an error about a timer and interpreted it as timers not working for compiletime: `Variable GameTimer_gameTimer is not initialized`
So I guess the real issue is that the timer is initialized inside an init block, which is not run for unit tests?

Frotty:
I see, if it's defined in place it works, but get elapsed isn't defined, will add.

Frotty:
Ok I pushed commit on wurst and stdlib, test should be possible now.

Jampi0n:
The class also uses timer.getRemaining() and something with the compiletime implementation of get elapsed is wrong. I get values in the magnitude of 1.6*10^9.
I think the issue is that the timer is only started if there is a function to execute, because I get reasonable values if I change the function. It also should return real numbers. Currently I only get integer numbers, so it will mostly be rounded down to 0 unless you have a really long unit test.

On a different note, should additional imports be avoided for unit tests? For example I could make the unit test more readable with LinkedList and Printing, but they would only be used by the unit test.

Frotty:
Okay maybe need to check implementations then, a simple unit test should suffice.

PR #394, opened 2021-07-16
Group API

GetLocalPlayer:
API changes:

```
.addUnit(unit) -> .add(vararg unit)
.addGroup(group) -> .add(vararg group)
.removeUnit(unit) -> .remove(vararg unit)
.removeGroup(group) -> .remove(vararg group)
.contains(unit) -> .has(unit)
.getUnitAtIndex(int index) -> .get(int index)
.getRandomUnit() -> .getRandom() // the body is rewritten with .get
```

Cosmetic:
```
boolean -> bool
integer -> int
```

### DIFF
--- a/wurst/_handles/Group.wurst
+++ b/wurst/_handles/Group.wurst
@@ -74,7 +74,7 @@ public function group.destr()
 public function group.has(unit u) returns bool
 	return IsUnitInGroup(u, this)
 
-@deprecated("Use .has insted.")
+@deprecated("Use .has instead.")
 public function group.contains(unit u) returns bool
 	return this.has(u)
 
@@ -92,7 +92,7 @@ public function group.getRandom() returns unit
 		return this.get(GetRandomInt(0, size - 1))
 	return null
 
-@deprecated("Use .getRandom insted.")
+@deprecated("Use .getRandom instead.")
 public function group.getRandomUnit() returns unit
 	return this.getRandom()
 
--- a/wurst/_handles/Group.wurst
+++ b/wurst/_handles/Group.wurst
@@ -2,6 +2,7 @@ package Group
 import NoWurst
 import Vectors
 import Player
+import Annotations
 
 /** Use this group for your non-nested group enum calls
 	and then iterate through the group via **for from** or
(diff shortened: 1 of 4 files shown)


wurst/_handles/Group.wurst
```diff
@@ -35,86 +36,123 @@ public function group.enumUnitsOfType(int id, boolexpr filter)
...
+		this.remove(u)
 
 public function group.destr()
 	DestroyGroup(this)
 
-public function group.contains(unit u) returns boolean
+public function group.has(unit u) returns bool
 	return IsUnitInGroup(u, this)
 
+@deprecated("Use .has insted.")
```
Frotty, line 77: instead*

wurst/_handles/Group.wurst
```diff
@@ -35,86 +36,123 @@ public function group.enumUnitsOfType(int id, boolexpr filter)
...
-	ForGroup(this) ->
-		randomCounter--
-		if randomCounter == 0
-			randomSelected = GetEnumUnit()
+	if size > 0
+		return this.get(GetRandomInt(0, size - 1))
+	return null
 
-	return randomSelected
+@deprecated("Use .getRandom insted.")
```
Frotty, line 95: instead*

wurst/_handles/GroupTests.wurst
```diff
@@ -8,36 +8,39 @@ function testGroupBasic()
...
+
+	var count = 0
+
 	for i = 1 to 10
-		g.addUnit(createUnit(Player(0), 'hfoo', vec2(0, 0), 0 .fromDeg()))
+		count += g.add(createUnit(Player(0), 'hfoo', vec2(0, 0), 0 .fromDeg()))
 
 	assertTrue(g.size() == 11)
-
+	assertTrue(count == 10)
```
Frotty, line 20: should use assertEquals for int assertions

wurst/_handles/GroupTests.wurst
```diff
@@ -82,27 +85,65 @@ function testHandleId()
...
+	unit array units
+	var _count = 0
+
+	for i = 0 to 9
+		units[i] = createUnit(Player(0), 'hfoo', vec2(0, 0), 0 .fromDeg())
+
+	g.clear()
+
+	_count = g.add(units[0], units[1], units[2], units[3])
+	assertTrue(_count == 4)
```
Frotty, line 100: same for these assertions

Frotty:
Nice, thanks for the PR 👍 . I added a few comments.

GetLocalPlayer:
Done.
Must `assertTrue(count = g.size())` also be done through assertEquals?

Frotty:
> Must assertTrue(count = g.size()) also be done through assertEquals?

It is better, because then, in case of an error, the message can display actual and expected value, where for the boolean assertion it can only show true/false which isn't helpful in this case.

I wonder why CircleCI isn't building this branch..

GetLocalPlayer:
The rest are also replaced, except `.getHandeId()` comparisons.

Frotty:
Will merge now to see if it builds 😬

Frotty:
Nice, it does 👍

Issue #396, opened 2021-07-21
Is `.progress()` method acceptable for `CallbackCounted`?


GetLocalPlayer:
In timer closures you often need to have a progress info but `CallbackCounted` doesn't contain any except `.isLast()`.

Something like

```
public abstract class CallbackCounted
	private var count = 0
	private var remaining = 0
	private timer t

	protected abstract function call(thistype cb)

	protected function start(timer whichTimer, real time, int callAmount)
		count = callAmount
		remaining = callAmount
		/* ... */

	function progress() returns real
		return 1. - remaining / count
```
So `count` becomes the total number of calls. `.progress()` returns a value between 0 and 1.

Frotty:
Yea, seems reasonable 👍 If you wanna PR this, please also add a unit test.

GetLocalPlayer:
#397

PR #397, opened 2021-07-23
`.progress()` method for CallbackCounted

GetLocalPlayer:
#396

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,4 +9,4 @@ The standard library has greatly flourished due to amazing feedback, pull reques
 
 If further natives need to be implemented for compiletime, they can be requested via a Feature Request.
 
-Feel free to join our irc channel if you would like to contribute.
+Feel free to join our [Discord](https://discord.gg/mSHZpWcadz) if you would like to contribute.
--- a/wurst/_handles/Group.wurst
+++ b/wurst/_handles/Group.wurst
@@ -74,7 +74,7 @@ public function group.destr()
 public function group.has(unit u) returns bool
 	return IsUnitInGroup(u, this)
 
-@deprecated("Use .has insted.")
+@deprecated("Use .has instead.")
 public function group.contains(unit u) returns bool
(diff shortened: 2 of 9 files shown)


Frotty:
Nice 👍  But I think `count` should stay the way it was as to not break anything. Please just make `remaining` save the total call amount and leave other things unchanged.

GetLocalPlayer:
Is it possible, to break anything? `count` is a private field that's not accessable from anywhere except the class it contains as far as I know. Even from children.

Frotty:
Just to be sure and to keep changed to a minimum. There is no benefit of these changes, is there.

GetLocalPlayer:
What about this? I'm placing `count` decreasement before the callback execution, otherwise `.progress()` returns a wrong value (that's > 0) before the callback very first execution.

Frotty:
I don't think that's good, now count values will be different. The original behavior must not be changed by this PR.

Frotty:
It seems fine now, please add some assertions for the callback before the first invocation.

GetLocalPlayer:
> add some assertions for the callback before the first invocation

Like the last two lines of code?

Frotty:
Missed that on phone 👍 
Can you please fix the circleci build?
Check this: https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories
Fork builds are enabled on my side, seems like you have to follow the second note:

>Note: If a user submits a pull request to your repository from a fork, but no pipeline is triggered, then the user most likely is following a project fork on their personal account rather than the project itself of CircleCi, causing the jobs to trigger under the user’s personal account and not the organization account. To resolve this issue, have the user unfollow their fork of the project on CircleCI and instead follow the source project. This will trigger their jobs to run under the organization when they submit pull requests.

GetLocalPlayer:
Fixed?

Frotty:
I see, your fork isn't up to date with upstream, consider merging. Looks good now, thanks 👍

PR #400, opened 2021-07-30
updated framehandles

Krypt0n:
- added real whole screen coord for frames
- added settext and settext per players basis
- added world ui frame


Frotty:
Looks good, thanks 👍

PR #401, opened 2021-08-03
Framehandles settings null player

Krypt0n:
@Donach  review?

### DIFF
--- a/wurst/_handles/Framehandle.wurst
+++ b/wurst/_handles/Framehandle.wurst
@@ -116,12 +116,12 @@ public function createFrame(string name) returns framehandle
 		name: the name of the frame which must be defined in the imported fdf files.
 		owner: the parent of the frame
 		priority: unknown
-		createContext: the unique id assigned to the frame and its children (useful for creating multiple frames of the same name) */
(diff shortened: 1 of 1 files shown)


Donach (approved):
All seems ok to me.

Frotty:
Thanks for the PR 👍 
It does indeed look ok, but I'm not a fan of the non-functional dot changes spamming the diff without providing any real benefit.
Rather keep them out or in another, solely format focused PR, since now I don't actually know what the critical change is.

Krypt0n:
Okay Ii can remove that, someone added the dots there in some other commit so I ditched to the same format
Frotty really... "solely format focused PR, since now I don't actually know what the critical change is." such a bullshit comment is really mean. Thought know it better, how disappointing ...

Sometimes you comply about meticulous things like this, then if someone takes an effort for those, you decide randomly to the opposite.

Frotty:
I'm sorry you feel that way, there was no such intent in my message.
I tried phrasing it as neutral as possible and refrain from personal remarks, I would prefer it if you do the same.
Also I didn't say I would not merge the PR - I explained why these changes make reviewing the PR more difficult.
Things exactly like this have caused failing/broken builds after PRs were merged before - formatting changes intermixed with functional changes, resulting in overlooked lines.

PR #402, opened 2021-08-19
Missing natives and some hot docs

GetLocalPlayer:
`EventData`

```

	/** Returns `true` if the damage done is done with Unit's attack. */
	static function isAttackDamage() returns bool
		return BlzGetEventIsAttack()
```

`Unit`

````

public function unit.startAbilityCooldown(int abilCode)
	BlzStartUnitAbilityCooldown(this, abilCode, this.getAbilityCooldown(abilCode, this.getAbilityLevel(abilCode) - 1))

public function unit.startAbilityCooldown(int abilCode, real cooldown)
	BlzStartUnitAbilityCooldown(this, abilCode, cooldown)

```

### DIFF
--- a/wurst/_handles/Unit.wurst
+++ b/wurst/_handles/Unit.wurst
@@ -720,12 +720,12 @@ public function unit.getCollisionSize() returns real
 	return BlzGetUnitCollisionSize(this)
 
 /** Level is indexed. Level 1 = 0, Level 2 = 1, e.t.c. */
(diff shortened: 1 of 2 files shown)


GetLocalPlayer:
Changed `level` param name to more obvious `levelIndex`

Frotty:
Sry I missed this somehow, thanks for the PR!

PR #403, opened 2021-09-22
Fix isOlderThan and isNewerThan are backwards

Cokemonkey11:


Frotty:
unit test?

Cokemonkey11:
@Frotty thanks for looking - fixed

PR #405, opened 2021-10-25
Function module

Cokemonkey11:


Cokemonkey11:
Having given this some further thought, I think it only makes sense to merge if there's a path forward to getting internal usage of this package in the stdlib

Given that we can't do that, it doesn't seem fair to recommend this in the stdlib

I do think this points to a bit of a flaw though - we should be able to do this without perf impact. I consider this cleaner than the current status quo

PR #407, opened 2021-11-13
Public import for Icons.wurst in ChannelAbilityPreset

Overkane:
If you use channel ability preset package, then you will most likely want to set an icon, since you don't wanna use std one.

This change will remove need in redundant Icons.wurst imports in every ability package, which use ChannelAbilityPreset.


Frotty:
The main reason we don't auto-import the asset packages is because of their size and thus longer compile time.
I think the icons import is small enough and often used to have it included - especially after the typecheck improvements.

PR #409, opened 2021-12-09
gaussian rng

Cokemonkey11:
Inlines https://www.hiveworkshop.com/threads/logarithm.247774/

Uses https://en.m.wikipedia.org/wiki/Box%E2%80%93Muller_transform

See also https://www.hiveworkshop.com/threads/standard-deviation-random-number.336870/

### DIFF
--- a/wurst/math/Maths.wurst
+++ b/wurst/math/Maths.wurst
@@ -11,8 +11,16 @@ public constant real EULER	  =  2.718281828
 public constant real PI			=  3.141592654
 /** PI * 2 constant */
 public constant real PI2 		=  6.28318
+/** One turn in radians */
+public constant real TAU		=  PI2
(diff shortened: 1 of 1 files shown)


Frotty:
I think it's debatable whether the rng is stdlib material, but log seems useful.
I don't understand the redeclarations of tau and max_real.
How about unit test?

Cokemonkey11:
@Frotty I would be happy to inline TAU and MAX_REAL if you'd like. They seem broadly useful missing constants to me?

For unit test I can add one for log. I previously wrote one just to print RNG to console but I'm not sure what a good unit test looks like that won't flake. Do you have any suggestions?

Frotty:
> They seem broadly useful missing constants to me?

They aren't missing though?

PR #411, opened 2022-03-13
Added getUnitIcon method

Jaccouille:
I didn't know it was possible to retrieve a unit icon, then I saw [this hive workshop comment](https://www.hiveworkshop.com/threads/still-no-natives-to-get-unit-type-icons-etc.337444/#post-3516994), I thought adding this method would be useful, at least to know that it's possible.


Frotty:
Thanks for the PR, tasyen says "One can get the icon of unit/item/ability" so I suppose it does more than just units, maybe that should be reflected in the base function?

PR #412, opened 2022-04-12
Features added

HerlySQR:
I'm not sure if this is well done.


HerlySQR:
In case you read the original message, sorry for that I though I was requesting to myself.
And sorry for being ignorant, I'm very noob in git, but what means the previous warn message?

Frotty:
I think these are too many changes to review. It would be nice to merge this, but I need some way to verify this. CircleCI fails because the configuration is missing. you should also add that from master.

HerlySQR:
I don't completly understand, what can I do?

HerlySQR:
I lost the thread, what could I do?

Frotty:
Hi, these are too many changes for me to review.
If you could fix the CircleCI build (https://app.circleci.com/pipelines/github/wurstscript/WurstStdlib2/624/workflows/46c70057-1fb4-4bfc-b4f5-5946b641e11e/jobs/646) then we at least could verify that everything you added compiles, even if I didn't test it.
Could you please fix the pipeline?

HerlySQR:
Now that the check have passed, what will happen?

Frotty:
Looking good. The only thing I think is missing is that you don't provide the 1.29 common.j and blizzard.j, so the compilation is done with the latest files, which could bring some potential errors. Did you test this with 1.29 files?

HerlySQR:
> Looking good. The only thing I think is missing is that you don't provide the 1.29 common.j and blizzard.j, so the compilation is done with the latest files, which could bring some potential errors. Did you test this with 1.29 files?

Yes it worked with it, I even outdated them, but when I got the error of `GetBJPlayerNeutralVictim` I updated them again and didn't oudated it because I didn't know what would affect it.

HerlySQR:
Should I outdate it again?, because anyway it tells me "Is not part of the repository".

Frotty:
I'm not sure what you mean. The branch is supposed to work with 1.28 and lower versions. So you could add 1.28 common.j and blizzard.j to the repo and it should be used instead of 1.32

HerlySQR:
So it's still ok?

HerlySQR:
I hope the blizzard.j and common.j were added correctly.

Frotty:
When I tested locally the wrong common.j / blizzard.j are used in vscode, so I'm not sure if it works as expected. Also not sure if it would work properly as dependency with those files commited. I guess in the end we need proper version selection in wurst.build or so. I guess I will merge this without .j files for now.

HerlySQR:
Soooo that means you will merge my commits?

HerlySQR:
I didn't wanna say this, but I can't continue making my map if this library is not added, because I don't wanna screw it again.

Frotty:
How so?
You can simply supply the git repository of your fork in `wurst.build` instead.

```
---
projectName: Zombie Defense
dependencies:
- https://github.com/HerlySQR/WurstStdlib2:pre1.29
```

HerlySQR:
> 

As I said, I'm very noob in Git, thank you.

HerlySQR:
Hmm, is not working, I add the link in the dependencies and in the `wurst build` and run `grill install` but the library is still the base, even I created a new project with that link in the dependencies, and still not.

PR #413, opened 2022-07-14
Add tests for mocked Quest and Quest Item handles.

thecadorn:
Add tests for the quest and quest item handles, leveraging the mocks added in https://github.com/wurstscript/WurstScript/pull/1060. Prefers testing through the Wurstscript handle implementations over the natives if possible.

Tried to base these on existing tests and the style guide, feedback is of course welcome!


Frotty:
Sorry I must have missed this. Thanks a lot. 👍

PR #417, opened 2022-11-01
Damage Engine Fixed

HerlySQR:
Hello, I made the changes of the Damage Engine, but I'm not sure of how just not show all the commits, but I think now just show the last changes.


wurst/_handles/Destructable.wurst
```diff
@@ -1,8 +1,11 @@
 package Destructable
 import NoWurst
-import Vectors
 import Wurstunit
 import Annotations
+import initlater Unit
```
Frotty, line 5: don't add initlater import. Also this stuff does not look like it belongs in stdlib. Put it into some own package and then you also don't need initlater.

HerlySQR: I could add a package called ElevatorWalls, but where could be good place it?

Frotty: In your own project, not the stdlib.

HerlySQR: ok

HerlySQR: Very well, I removed it.

Frotty:
Looking good now, thanks 👍

PR #421, opened 2023-05-08
SHIFTR operator negative

Elphea:
Fixed negative numbers when pertaining to SHIFTR operator

This version allows usages between -2^31 to 2^31-1.


Frotty:
Addressed by https://github.com/wurstscript/WurstStdlib2/pull/430
The reason for the look of the function with just 1 return is, so that it can be inlined.

PR #423, opened 2023-06-24
Added unit showTeamGlow native

Jaccouille:
Used to remove hero glow on a unit.
I wonder what exactly is considered a glow though.


Frotty:
Thanks 👍

PR #424, opened 2023-07-02
Fixed Grom Hellscream soundset path

Jaccouille:
On 1.36, "Hellscream" doesn't seem to exist.

I didn't look for the reason as to why, but I remember some odd change about Hellscream name for reforged, maybe it's related?
https://us.forums.blizzard.com/en/warcraft3/t/why-do-we-hate-grom-now/14474


Frotty:
Thanks

PR #425, opened 2023-07-22
Added pathing map assets

Jaccouille:
Related to #180 
I took the pathing texture data from the unit & destructable SLK file.

I chose to have `PathingMaps.simpleSolid4x4` instead of `PathingMaps.farm` as described in the issue.
My personal opinion, it's more explicit and avoids duplicates pointing to the same texture path.

I based the constant name on the filename, which might gives weird results like `stoneWall4Path`.


Frotty:
Sorry, forgot to get to this. Thanks.

PR #429, opened 2024-01-02
Expose timer `getRemaining` API in `CallbackSingle`

jmrico01:
I implemented a spell that has multiple charges, and an ability cooldown timer is displayed for each recharge time. I'm using `doAfter` to defer the recharge event, but I also need to update the remaining cooldown. All that to say, I need `CallbackSingle` to expose the `getRemaining()` API for its underlying timer.

Realized after my last PR that I've been using my WurstStdlib2 fork this whole time because of this 2-line change, so I figured I'd submit a PR for it.


Frotty:
Thanks for the submission 👍

PR #431, opened 2024-01-30
Added unit method for item stock handling

Jaccouille:
Wrapping up some native functions related to item stock.


Frotty:
Thanks 👍

PR #432, opened 2024-01-30
Ability assets correction

Jaccouille:
* Added/fixed few ids (heal reduction, sell item, spell steal)
* Added sundering blade model path
* Added AbilityDefinitionHealReductionBonus
* Added thunderclap setter for maximum damage


Frotty:
Very nice 👍 Did `sellItems` change in reforged or was it always wrong?

Jaccouille:
Based on the old version I have on my hard drive (checked units/abilitydata.slk), it was right on 1.28.5, wrong on 1.31 (24/04/2019)

PR #436, opened 2024-04-03
Added a few missing assets.

Jaccouille:
Added a few missing assets.
Why does the soundSets.wurst file contains comments like `// RazorMane` without the sound set string?


Frotty:
Thanks. Usually, comments were only made for identical assets used by multiple units. It seems that some are missing/incorrect though.

PR #439, opened 2024-12-08
Added setPortraitModelFile method

Jaccouille:
The "upor" or "Art - Portrait Model File" field was added in the patch 2.0, it's a field used to set the portrait model, I believe its purpose is related to the new skin feature.

From what I am seeing, it's set to default for most standard editor units, if you make a new custom unit and change its model, you're going to have to change this portrait field too.


Frotty:
Thanks, yea it needs to be changed with 2.0 now, wanted to add it as well.

PR #440, opened 2024-12-11
Added Environment folder assets

Jaccouille:
Contains path to model files like the burning effect on damaged buildings.


Frotty:
Thanks. Can you please add it to the Assets package as well?

PR #441, opened 2025-02-05
Add support to DummyCaster package for vec3 origins

thecadorn:
Following conversation in the discord, support vec3 origins. Keep support for vec2 origins and convert them internally.


Frotty:
Thanks for the PR. However, it seems that you are not using the z component of origin anywhere?
In line 86 it's still just doing `setXY`.

thecadorn:
Oh no, I was definitely rushing through this. Thanks for catching it!

Updated with setXYZ. It looks like this one sets the position relative to the terrain height, whereas setXYZReal sets an absolute position.

Frotty:
Thanks.

> Updated with setXYZ. It looks like this one sets the position relative to the terrain height, whereas setXYZReal sets an absolute position.

Yep, should be the correct choice when doing dummy casts for air units.

PR #452, opened 2026-07-25
feat(stdlib): generic fixes + helpers from Twilight's Eve Reborn

Donach:
## Summary

A batch of small, generic stdlib improvements and additions extracted from the **Twilight's Eve Reborn** (TeveR) project, which maintained a WurstStdlib2 fork for years. TeveR's SQLite compiletime work recently landed upstream (WurstScript #1180); this contributes the remaining genuinely-generic pieces so the fork can be fully retired. Everything here is in production use in that map.

Grouped by kind:

### Bug fixes / hardening
- **`LinkedList` — null-safe `LLIterator.reset()`** (both iterator classes): `reset()` is called from the constructors; if an iterator is ever built with a `null` parent it null-crashes. Guarded with `parent != null ? parent.getDummy() : null`. *(Supersedes #451, folded in here at the maintainers' preference for a single PR.)*
- **`SyncSimple` — null `syncBuffer` guard**: on the buffered path `syncBuffer` was observed null under Campaign Edition / replay, null-crashing the whole sync queue on a late/duplicate sync event. Recreate it defensively before append.
- **`EffectUtils` — null-guard `destrAfter` / `destrHiddenAfter`**: the deferred `doAfter` closure could fire after the effect was already destroyed/null.

### Additions
- **`ClosureTimers.CallbackPeriodic.enable()/disable()`**: pause/resume a periodic callback without destroying and rebuilding it.
- **`Integer.toStringOrdinal()`**: `1 -> "1st"`, `2 -> "2nd"`, `3 -> "3rd"`, … with the 11/12/13 exceptions handled.
- **`Unit.disableAttackEx(index, flag)`**: toggle a single weapon by index via the weapon field (unlike `disableAttack`, works per-weapon on multi-weapon units and doesn't hide the attack command).
- **`GroupUtils.group.getClosestUnit(vec2)`** and **`group.filterAliveUnits(mustBeEnemyOf, ownerId)`**.
- **`Rect`**: `rect.toRegion()` (ownership documented), `rect.fromCenter(...)`, `rect.fromCenterWithOffset(...)`.
- **`Vectors`**: 2D line/geometry helpers — `lineFormular` tuple + `getYOnLine`/`getXOnLine`/`getLineFormular`, parallel/orthogonal offsetting, `vec2.isCrossingLine`, and a `fourPointTuple` quad with `isPointInside` + bounds accessors. Includes `@Test` coverage.
- **`SyncSimple`**: generic `player.sync<T>(...)` + `AnySyncListener<T>` for syncing any int-representable value (class instances / handle ids already in sync across clients).

## Notes
- Additions carry doc comments and, where they allocate handles, explicit caller-ownership warnings.
- Happy to split this into separate PRs if the maintainers would prefer smaller units — bundled per request on our side.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/wurst/data/LinkedList.wurst
+++ b/wurst/data/LinkedList.wurst
@@ -4,6 +4,7 @@ import TypeCasting
 import ClosureForGroups
 // re-exports Integer, String, Real and Printing
 import ErrorHandling
+import Wurstunit
 
 /**
  * Doubly-linked list implementation that implements all common list, stack and queue operations.
@@ -464,10 +465,12 @@ public class LLIterator<T>
 	protected var destroyOnClose = true
 
 	construct(LinkedList<T> parent)
+		assertTrue(parent != null)
 		this.parent  = parent
 		reset()
 
 	construct(LinkedList<T> parent, bool destroyOnClose)
+		assertTrue(parent != null)
 		this.parent  = parent
 		reset()
 		this.destroyOnClose = destroyOnClose
@@ -527,11 +530,13 @@ public class LLBackIterator<T>
 	protected bool destroyOnClose = true
 
 	construct(LinkedList<T> parent, bool destroyOnClose)
+		assertTrue(parent != null)
 		this.parent = parent
 		reset()
 		this.destroyOnClose = destroyOnClose
 
 	construct(LinkedList<T> parent)
+		assertTrue(parent != null)
 		this.parent = parent
 		reset()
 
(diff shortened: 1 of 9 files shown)


wurst/data/LinkedList.wurst
```diff
@@ -451,7 +451,7 @@ public class LLIterator<T>
 		this.destroyOnClose = destroyOnClose
 
 	function reset()
-		this.dummy   = parent.getDummy()
+		this.dummy   = parent != null ? parent.getDummy() : null
```
Frotty: parent should never be null? Would be better to assert it not being null during construct imo, because that would already be a failure state? Unless I'm missing a legit case.

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

PR #454, opened 2026-08-01
feat: add reusable Polygon abstraction

Donach:
## Summary

- add an arbitrary-size `Polygon` class with builder and vararg construction
- cache and own the polygon boundary `rect` after sealing
- accelerate exact classification with a 32-world-unit pathing-cell grid: uniform cells answer in O(1), mixed cells inspect only their horizontal-band edge candidates, bounded by O(n)
- expose O(1) `vec2` / `vec3` to global `pathingCell` conversion and cell bounds/center
- keep `classifyLinear` beside the accelerated `classify` as the exact reference and safe fallback
- store grid data in a reusable 32-array sharded integer arena, avoiding the single JASS-array limit
- provide exact even-odd behavior for convex, concave, degenerate, reversed, and self-intersecting paths
- add overridable debug rendering with optional color, caller-owned handles, and timed automatic cleanup
- retain and deprecate the existing `vec2.isInPolygon` and `vec3.isInPolygon2d` APIs for source compatibility

## Cell semantics

A cell is `INSIDE` only when its entire area is inside without touching an edge, `OUTSIDE` only when disjoint from the polygon, and `MIXED` when any polygon edge touches or crosses it. `classify` resolves `MIXED` cells exactly.

## 10,000-query evaluation

The accelerated and linear implementations returned identical relations for all 10,000 deterministic points against a 64-edge polygon. The linear path considered 506,048 edges for in-bounds queries; the accelerated path considered 22,486 candidate edges (about 22.5x fewer). 1,978 queries needed edge candidates, so 8,022 were answered without an exact edge scan.

## Ownership

`Polygon` owns its cached bounds and grid allocation. Untimed debug rendering returns a caller-owned `ArrayList<lightning>`; timed rendering owns and cleans its handles.

## Verification

Run inside `nix develop /home/daniel/Repositories/tever/tever_main`:

- `grill typecheck --quiet` — passed
- `grill test PathingGridTests` — 3/3 passed
- `grill test ShardedIntStorageTests` — 3/3 passed
- `grill test PolygonTests` — 20/20 passed
- `grill test testIsInPolygon` — 1/1 passed
- `grill test --quiet` — passed
- GitHub Actions `Build` — passed
- `git diff --check upstream/master...HEAD` — passed

The non-quiet compiler output intentionally reports six deprecation warnings from the restored legacy API test. Warcraft-native lightning creation, coloring, and timed destruction are compile-checked but not Warcraft-runtime verified because Grill cannot execute those natives.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Update Wurst
         run: grill install wurstscript
@@ -21,4 +21,4 @@ jobs:
         run: grill install
 
       - name: Test
-        run: grill test
+        run: grill test --quiet
--- a/wurst/data/ShardedIntStorage.wurst
+++ b/wurst/data/ShardedIntStorage.wurst
@@ -1,0 +1,250 @@
+package ShardedIntStorage
+import NoWurst
+import ErrorHandling
+import Table
+
+/**
+	A shared integer arena spanning multiple native JASS arrays. Allocations are
+	contiguous logical address ranges and released adjacent ranges are coalesced.
+*/
+public class ShardedIntStorage
+	private static constant int SHARD_COUNT = 32
+
+	private static int array storage0
+	private static int array storage1
+	private static int array storage2
+	private static int array storage3
+	private static int array storage4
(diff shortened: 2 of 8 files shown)


Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

PR #464, opened 2026-08-12
Deprecate legacy serialization entrypoints

Frotty:
## Summary
- deprecate the legacy Serializable serialize/deserialize entrypoints with a migration hint
- document why structured serialization uses a strict varint wire alphabet instead of the general-purpose Base64 codec

## Verification
- grill typecheck
- grill test --quiet
- git diff --check


Frotty:
@codex review

PR #465, opened 2026-08-12
Add specialized IntMap for structured serialization

Frotty:
## Summary

- add IntMap with direct integer keys and compiler-specialized typed value storage
- replace the structured reader's legacy HashMap and IterableMap pair with one typed field map
- make migration replacement and rename ownership explicit and correct
- cover primitive, class, tuple, replacement, removal, clear/reuse, and reader-mutation behavior

## Design

Wurst currently rejects mixed legacy/specialized generic declarations, while JASS hashtables require integer child keys. IntMap is therefore the smallest backend-aligned abstraction: the key needs no adapter and values avoid legacy integer erasure and typecasting.

## Verification

- grill typecheck --quiet
- grill test --quiet
- git diff --check

### DIFF
--- a/wurst/data/IntMap.wurst
+++ b/wurst/data/IntMap.wurst
@@ -1,0 +1,96 @@
+package IntMap
+import NoWurst
+import ArrayList
+import Table
+
+/**
+ * O(1) integer-keyed map with compiler-specialized value storage.
+ *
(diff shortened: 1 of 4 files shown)


Frotty:
@codex review

PR #468, opened 2026-08-17
Add a hash map which keeps its keys as they are

Frotty:
`HashMap` casts every key to an int and stores it in a `Table`. That works for handles and for anything castable, but it loses the type on the way in: two keys which cast to the same int collide, and a key which is not castable cannot be used at all.

`FastHashMap` takes a bound on its key type instead, so hashing and comparing are done by the key's own implementation and the key is stored as itself:

```wurst
implements Hashable<vec2>
    function hash(vec2 v) returns int
        return v.x.toInt() * 31 + v.y.toInt()
    function equals(vec2 a, vec2 b) returns boolean
        return a == b

let seen = new FastHashMap<vec2, unit>()
seen.put(caster.getPos(), caster)
```

Instances for `int` and `string` come with the package. Declaring one beside your own type is what makes it usable as a key.

### Storage

One array per specialisation, carved into a section per instance, the way `ArrayList` works. A section is `FASTHASHMAP_CAPACITY` slots and does not grow; collisions probe linearly inside it, and a removed slot becomes a tombstone rather than empty so a probe which passed over it still finds keys put down beyond it.

A destroyed map hands its section back, so `FASTHASHMAP_MAX_INSTANCES` bounds maps *alive* rather than maps ever made. Every section is the same width, so released ones are a stack rather than the capacity-matched free list `ArrayList` keeps — any released section fits any new map, and there is nothing to compact. On the Jass target the configured total is also checked against `JASS_MAX_ARRAY_SIZE`, the storage being one fixed-size array.

### The hash is computed here

Neither instance takes a shortcut, because a section fills up: colliding keys do not merely probe longer, the map eventually refuses them.

`StringHash` is unusable for this. It is case insensitive, so `alpha` and `ALPHA` would share a slot, and it collapses every partial multibyte slice to one constant. The string hash mixes each byte with its position and the length instead, so anagrams and prefixes separate too. Single bytes still decode through `StringUtils.char`, which recovers the case `StringHash` loses and which the library already depends on throughout; non-latin text is the remaining weak case, its lead bytes not decoding, and the doc says so.

The int hash returned the key itself at first, which sends everything strided by the capacity to slot zero — ids, handles and loop counters all arrive strided like that. Halves are mixed separately so the multiplications stay in range.

Every intermediate stays below 2^31 rather than relying on overflow, which wraps at 32 bits on Jass and does not on Lua. Nothing stores a hash, so differing values would not have been wrong, but the interpreter could then no longer stand in for the game while testing distribution.

### Reporting and iteration

A full map reports a dropped write rather than discarding it silently — losing a store without a word is close to impossible to find from the outside. A key the map already holds stays writable, and `isFull()` asks beforehand.

Entries can be walked without allocating:

```wurst
var slot = map.nextEntry(0)
while slot >= 0
    doSomething(map.keyAt(slot), map.valueAt(slot))
    slot = map.nextEntry(slot + 1)
```

Slot walking rather than an iterator object or a closure: nothing is allocated, and it avoids dispatching a bound through a closure, which is not supported on every target. A `for in` wrapper can be layered on this.

### Compiler requirement

Type class bounds for `T:` generics, shipped in wurstscript/WurstScript#1226, #1228 and #1229. The bound is what makes the key's own `hash` and `equals` reachable from inside the generic, and it lowers to a direct call to the instance function after specialisation, on both Jass and Lua.

### Testing

`grill test` reports 489/489, of which 25 are this package's. Beyond the obvious behaviour — put/get, replacement, missing keys reading as the value type's default, colliding keys, removal keeping later keys reachable, tombstone reuse, independent instances, two specialisations coexisting — the tests cover the three things that would otherwise pass quietly:

- **The hash itself**, reached through a `hashOf<T: Hashable>` helper. The behavioural tests pass with an identity hash and with `StringHash`, because probing separates colliding keys however badly they hash; only observing the hash catches it. Against the old ones these fail with all sixteen strided keys in one slot and with `alpha` and `ALPHA` hashing alike.
- **Section reuse**, by creating and destroying past the instance limit. Without it, that reports `out of sections`.
- **Iteration**, including that tombstones are skipped and do not end the walk early.

Every test here was checked to fail before it passed.

### Note

The container is also exercised from the compiler side in `FastHashMapTests` there — standalone and with the standard library in scope, on both targets — since it is the shape the bounds feature was built for. One gap worth stating: the standard-library-in-scope case is compiled but not executed on Lua, because the test runtime shim cannot initialise the library's own packages.

A later step could give Lua a native table instead of this probing, which would lift `FASTHASHMAP_CAPACITY` there; that is wurstscript/WurstScript#1255, and it is deliberately not part of this.

### DIFF
--- a/wurst/data/FastHashMap.wurst
+++ b/wurst/data/FastHashMap.wurst
@@ -2,39 +2,6 @@ package FastHashMap
 import NoWurst
 import Wurst
 
-/**	A hash map which keeps its keys as they are.
-
-	`HashMap` casts every key to an int and stores it in a `Table`, which works for
-	handles and for anything castable but loses the type on the way in: two keys which
-	cast to the same int collide, and a key which is not castable cannot be used at all.
-	This map takes a bound on its key type instead, so hashing and comparing are done by
-	the key's own implementation and the key is stored as itself.
-
-	The bound is `Hashable`, which asks for a hash and an equality:
-
-		implements Hashable<vec2>
-			function hash(vec2 v) returns int
-				return v.x.toInt() * 31 + v.y.toInt()
-			function equals(vec2 a, vec2 b) returns boolean
-				return a == b
-
-		let seen = new FastHashMap<vec2, unit>()
-		seen.put(caster.getPos(), caster)
-
-	Instances for `int` and `string` come with this package. Declare one beside your own
-	type to use it as a key.
-
-	Storage is one array per specialisation, carved into a section per instance, the way
-	`ArrayList` works. A section is `CAPACITY` slots and does not grow, so a map which
-	fills up refuses further keys rather than rehashing - see `isFull`. Raise
-	`FastHashMap_CAPACITY` in your build config if you need larger maps; every map of one
-	key and value type pays that size.
-
-	Collisions are handled by linear probing inside the section. A removed slot becomes a
-	tombstone rather than empty, so a probe which passed over it still finds keys put down
-	beyond it.
-*/
-
 /**	What a key type has to provide to be used in a `FastHashMap`. */
 public interface Hashable<T:>
 	/**	Any int; keys which are equal must hash alike, or a lookup will miss them. */
@@ -63,6 +30,42 @@ implements Hashable<string>
 
 constant SLOTS = FASTHASHMAP_CAPACITY * FASTHASHMAP_MAX_INSTANCES
 
+/**	A hash map which keeps its keys as they are.
+
+	`HashMap` casts every key to an int and stores it in a `Table`, which works for
+	handles and for anything castable but loses the type on the way in: two keys which
+	cast to the same int collide, and a key which is not castable cannot be used at all.
+	This map takes a bound on its key type instead, so hashing and comparing are done by
+	the key's own implementation and the key is stored as itself.
+
+	The bound is `Hashable`, which asks for a hash and an equality:
+
+		implements Hashable<vec2>
+			function hash(vec2 v) returns int
+				return v.x.toInt() * 31 + v.y.toInt()
+			function equals(vec2 a, vec2 b) returns boolean
+				return a == b
+
+		let seen = new FastHashMap<vec2, unit>()
+		seen.put(caster.getPos(), caster)
+
+	Instances for `int` and `string` come with this package. Declare one beside your own
+	type to use it as a key.
+
+	Storage is one array per specialisation, carved into a section per instance, the way
+	`ArrayList` works. A section is `FASTHASHMAP_CAPACITY` slots and does not grow, so a map
+	which fills up refuses further keys rather than rehashing - see `isFull`. That and
+	`FASTHASHMAP_MAX_INSTANCES` are both configurable, and every map of one key and value
+	type pays the section size.
+
+	On the Jass target their product is bounded by `JASS_MAX_ARRAY_SIZE` as well, the storage
+	being one fixed-size array: a construction which would hand out slots past its end errors
+	instead. Lua grows the table, so only the section count applies there.
+
+	Collisions are handled by linear probing inside the section. A removed slot becomes a
+	tombstone rather than empty, so a probe which passed over it still finds keys put down
+	beyond it.
+*/
 public class FastHashMap<K: Hashable, V:>
 	private static K array keys
 	private static V array values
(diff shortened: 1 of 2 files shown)


Frotty:
CI is green now. Three rounds, and both review findings are in.

**Build failure (mine).** The overview comment sat at package level with another doc comment after it, so it documented nothing and hotdoc rejected the position. Moved onto the class it describes, which is where `LinkedList` keeps its own. A second, masked error surfaced once that was fixed: `Wurst` re-exports `MagicFunctions`, which is where `isLua` comes from, but not `ErrorHandling`, so both `error()` calls resolved to nothing — the earlier failure had stopped compilation before name resolution ran.

**P1, storage past the array bound — accepted.** The section count was checked against the configured total and not against the array holding it, so raising `FASTHASHMAP_CAPACITY` while leaving `FASTHASHMAP_MAX_INSTANCES` alone — which the overview suggested doing — put that total past `JASS_MAX_ARRAY_SIZE` and handed out sections past the end of a fixed-size array. Guarded as `ArrayList.allocateStorage` guards it, on the Jass target only since Lua grows the table. The overview says so now, and no longer names a constant that does not exist.

**P2, retention on Lua — accepted.** A vacated slot kept the key and value it held, so a removed entry and a cleared map held their last occupant for the lifetime of the specialisation. Released in both `remove` and `clear` behind the same `isLua` branch `ArrayList` uses for the same reason; a tombstone is never read for its key, so it is safe.

Verified against the whole library rather than pushed to find out this time: `grill test` reports **474/474**, with all fourteen of this package's test functions executing by name.

@codex review

Frotty:
Two rounds since: a real hash, and the lifecycle gap that would have bitten in an actual map.

**The hash is computed here now, not taken from `StringHash` or from identity.** Both instances leaned on something that collides on ordinary keys, and a section is `FASTHASHMAP_CAPACITY` slots wide and fills up — so colliding keys do not merely probe longer, the map eventually refuses them. The string hash mixes each byte with its position and the length; single bytes still decode through `StringUtils.char`, which recovers the case `StringHash` loses and which the library already depends on throughout. The int hash mixed nothing at all, so everything strided by the capacity went to slot zero — ids, handles and loop counters all arrive strided like that. Every intermediate stays under 2^31 rather than relying on overflow, which wraps at 32 bits on Jass and does not on Lua.

Correction to something I wrote earlier in this PR: the interpreter *does* emulate `StringHash` — `StringProvider` delegates to `Wc3StringHash`, with `Wc3StringHashTest` checking parity against the Lua shim. The case insensitivity and the multibyte collapse are real and are the reason not to use it, but my claim that a test could not see them was wrong.

**A destroyed map now hands its section back.** `nextFree` only ever grew and nothing was released, so `FASTHASHMAP_MAX_INSTANCES` counted maps ever made rather than maps alive: a map creating one per spell cast or per unit exhausted the sections and every later one failed. That is the ordinary way a map uses a container, so it had to go before this is usable in one. Sections are all the same width, so released ones are a stack rather than the capacity-matched free list `ArrayList` keeps — any released section fits any new map, and there is nothing to compact.

**On the tests.** The first batch I wrote for the hash passed against the old hashes too, because probing separates colliding keys however badly they hash — coverage in name only. Reaching the requirement through a `hashOf<T: Hashable>` helper is what observes the hash itself, and against the old ones those fail with all sixteen strided keys in one slot and with `alpha` and `ALPHA` hashing alike. Same discipline for the lifecycle tests: with the reuse disabled they report `out of sections`.

`grill test`: **485/485**, all 25 of this package's test functions.

@codex review

Frotty:
Both gaps closed. `grill test`: **489/489**.

**A dropped write reports.** `put` silently discarded a new key once the section was full — the map simply does not have what you put in it, which is close to impossible to find from the outside. It errors now. A key the map already holds stays writable, and `isFull()` is there to ask beforehand. `testFullMapRefusesNewKeys` relied on the silent drop, so it asks `isFull` instead of writing into a full map.

**A map can be walked.** There was no way to see what one holds:

```wurst
var slot = map.nextEntry(0)
while slot >= 0
    doSomething(map.keyAt(slot), map.valueAt(slot))
    slot = map.nextEntry(slot + 1)
```

Slot walking rather than an iterator object or a closure: it allocates nothing, which is what a map iterating every frame needs, and it avoids dispatching a bound through a closure — a shape that is not supported on every target, as `TypeClassTests.nestedClosuresInsideBoundedGenericAreRejectedForLua` pins. A `for in` wrapper can be layered on this later; this is the primitive underneath it.

`nextEntry` skips tombstones, so a removed key is not visited and its slot does not end the walk early. `keyAt`/`valueAt` report on a slot holding nothing rather than handing back whatever the array happens to have there. Four tests cover it: every entry once, removed entries skipped, empty and cleared maps, and string keys iterating as themselves.

One naming note for the next reader: the parameter is `startSlot` because `from` is reserved.

@codex review

Frotty:
Description rewritten — it still described only the original container, and the branch has since grown a custom hash, section reuse on destroy, a reported rather than silent dropped write, and iteration. Six commits is a lot to reconstruct that from.

One thing worth stating explicitly for review: this branch predates #467, so a tip-to-tip diff shows `PolygonTests.wurst` as changed. It is not — against the merge base the PR touches exactly two files, `FastHashMap.wurst` and `FastHashMapTests.wurst`, which is what the Files tab shows. Nothing here reverts the polygon fix.

@codex review

PR #471, opened 2026-08-31
Add preserveName annotation to magic functions

Frotty:
## Summary

- Add the public `@preserveName` annotation declaration to the compiler magic-functions package.
- Document that it preserves externally referenced top-level function names when `-opt` name minification is enabled.

## Checks

- `grill typecheck --quiet`

## Scope

This exposes the annotation to stdlib consumers; the compiler implementation is provided by the WurstScript compiler PR.


Frotty:
@codex review

